PR #134, opened 2014-09-27
Trace exception local variables and args per frame

mrap:
New configuration options and defaults
config.locals.enabled = off
config.locals.max_trace_frames = 100

Turned off by default because this may not be performant just yet!
1. Stores each Ruby process' local vars and method arguments
2. When gathering exception data, "replay" over each frame to get the variables for that frame's context
3. Dump the cache and repeat

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -253,8 +253,8 @@ module Rollbar
       @trace_store ||= []
     end
 
-    def add_trace_frame_locals(filename, lineno, locals, args)
-      trace_store << { :filename => filename, :lineno => lineno, :locals => locals, :args => args }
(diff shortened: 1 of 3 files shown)


coryvirok:
@rokob can you take this over? This would be awesome to have in the SDK!

guyzmo:
hey! 🙌 just a note to express my interest in this feature as well, being a pythonista doing some Ruby now, I'm surprised to see that such a thing is that problematic in Ruby. I'm sadly unable to help on that 😕

echan00:
Interested in this feature as well

rivkahstandig3636:

ArturMoczulski:
This is conflicting with current master and I believe local vars in frames have been implemented since. If this is still requested and valid I will reopen.

PR #197, opened 2015-01-08
refactor sidekiq handler

jjb:
* DRYed up the code by moving the error handling code into its own method that can be reused
* DRYed up the code by removing redundant invocations of `Sidekiq.configure_server do |config|`
* moved PARAM_BLACKLIST out of the global namespace

(the first one is what inspired me to do the refactoring in the first place, because in my project i need to change the behavior and currently have to do a monkeypatch which replicates all the code. after this refactoring, i can use `alias_method_chain`)

let me know what you think!

/cc @dadadadave

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -2,21 +2,25 @@
 
 PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
 
-if Sidekiq::VERSION < '3'
-  module Rollbar
-    class Sidekiq
-      def call(worker, msg, queue)
-        yield
-      rescue Exception => e
-        params = msg.reject{ |k| PARAM_BLACKLIST.include?(k) }
-        scope = { :request => { :params => params } }
+module Rollbar
+  class Sidekiq
+    def self.handle_exception(msg_or_context, e)
+      params = msg_or_context.reject{ |k| PARAM_BLACKLIST.include?(k) }
+      scope = { :request => { :params => params } }
 
-        Rollbar.scope(scope).error(e, :use_exception_level_filters => true)
-        raise
-      end
+      Rollbar.scope(scope).error(e, :use_exception_level_filters => true)
+    end
+
+    def call(worker, msg, queue)
+      yield
+    rescue Exception => e
+      Rollbar::Sidekiq.handle_exception(msg, e)
+      raise
     end
   end
+end
 
+if Sidekiq::VERSION < '3'
   Sidekiq.configure_server do |config|
     config.server_middleware do |chain|
       chain.add Rollbar::Sidekiq
@@ -25,10 +29,7 @@ if Sidekiq::VERSION < '3'
 else
   Sidekiq.configure_server do |config|
     config.error_handlers << Proc.new do |e, context|
-      params = context.reject{ |k| PARAM_BLACKLIST.include?(k) }
-      scope = { :request => { :params => params } }
-
-      Rollbar.scope(scope).error(e, :use_exception_level_filters => true)
+      Rollbar::Sidekiq.handle_exception(context, e)
     end
   end
 end
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -1,9 +1,9 @@
 # encoding: utf-8
 
-PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
-
 module Rollbar
   class Sidekiq
+    PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
+
     def self.handle_exception(msg_or_context, e)
       params = msg_or_context.reject{ |k| PARAM_BLACKLIST.include?(k) }
       scope = { :request => { :params => params } }
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -20,14 +20,12 @@ module Rollbar
   end
 end
 
-if Sidekiq::VERSION < '3'
-  Sidekiq.configure_server do |config|
+Sidekiq.configure_server do |config|
+  if Sidekiq::VERSION < '3'
     config.server_middleware do |chain|
       chain.add Rollbar::Sidekiq
     end
-  end
-else
-  Sidekiq.configure_server do |config|
+  else
     config.error_handlers << Proc.new do |e, context|
       Rollbar::Sidekiq.handle_exception(context, e)
     end
--- a/spec/rollbar/sidekiq_spec.rb
+++ b/spec/rollbar/sidekiq_spec.rb
@@ -1,0 +1,54 @@
+require 'spec_helper'
+
+unless RUBY_VERSION == '1.8.7'
+  require 'sidekiq'
+  require 'rollbar/sidekiq'
+end
+
+describe "Sidekiq integration", :reconfigure_notifier => false do
+  before{ skip if RUBY_VERSION == '1.8.7' }
+
+  describe "Rollbar::Sidekiq.handle_exception" do
+    let(:msg_or_context) { ["hello", "error_backtrace", "backtrace", "goodbye"] }
+    let(:exception) { StandardError.new("oh noes") }
+    let(:rollbar) { double }
+    subject do
+      Rollbar::Sidekiq.handle_exception(msg_or_context, exception)
+    end
+
+    it "constructs scope from filtered params" do
+      rollbar.stub(:error)
+      Rollbar.should_receive(:scope).with(
+        { :request => { :params => ["hello", "goodbye"] } }
+      ) {rollbar}
+      subject
+    end
+    it "sends the passed-in error to rollbar" do
+      Rollbar.stub(:scope) { rollbar }
+      rollbar.should_receive(:error).with(exception, :use_exception_level_filters => true)
+      subject
+    end
+  end
+
+  describe "middleware" do
+    let(:middleware) { Rollbar::Sidekiq.new }
+    let(:msg) { ["hello"] }
+    let(:exception) { StandardError.new("oh noes") }
+    subject do
+      middleware.call(nil, msg, nil) do
+        raise exception
+      end
+    end
+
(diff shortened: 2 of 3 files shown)


brianr:
Nice, thanks @jjb!

@jondeandres can you review this?

lib/rollbar/sidekiq.rb
```diff
@@ -1,36 +1,35 @@
...
     end
   end
+end
 
-  Sidekiq.configure_server do |config|
+Sidekiq.configure_server do |config|
+  if Sidekiq::VERSION < '3'
     config.server_middleware do |chain|
-      chain.add Rollbar::Sidekiq
+      chain.add Rollbar::SidekiqErrorHandlingMiddleware
```
jondeandres: I think you are expecting here to use `Rollbar::SidekiqErrorHandler`, which is the module name you've defined above.

lib/rollbar/sidekiq.rb
```diff
@@ -1,36 +1,35 @@
...
-  module Rollbar
-    class Sidekiq
-      def call(worker, msg, queue)
-        begin
-          yield
-        rescue Exception => e
-          params = msg.reject{ |k| PARAM_BLACKLIST.include?(k) }
-          scope = { :request => { :params => params } }
+  class SidekiqErrorHandler
+    def self.handle(scope, e, params)
```
jondeandres: This method should receive a `msg_or_context` and the `e` variables. Not `params` or `scope`, which are really the values `.handle` is calculating.

```ruby
def self.handle(msg_or_context, e)
  params = msg.reject{ |k| PARAM_BLACKLIST.include?(k) }
  scope = { :request => { :params => params } }

  Rollbar.scope(scope).error(e)
end
```
Does this make sense?

lib/rollbar/sidekiq.rb
```diff
@@ -1,36 +1,35 @@
...
-          raise
-        end
+      Rollbar.scope(scope).error(e)
+    end
+
+    def call(worker, msg, queue)
+      begin
+        yield
+      rescue Exception => e
+        SidekiqErrorHandler.handle(scope, e, params)
```
jondeandres: `scope` or `params` are not defined here. As I've commented above, this call should pass `msg` and `e` to `.handle`.

lib/rollbar/sidekiq.rb
```diff
@@ -1,36 +1,35 @@
...
-  end
-else
-  Sidekiq.configure_server do |config|
+  else
     config.error_handlers << Proc.new do |e, context|
-      params = context.reject{ |k| PARAM_BLACKLIST.include?(k) }
-      scope = { :request => { :params => params } }
-
-      Rollbar.scope(scope).error(e)
+      SidekiqErrorHandler.handle(scope, e, params)
```
jondeandres: `scope` or `params` are not defined here. As I've commented above, this call should pass `context` and `e` to `.handle`.

jjb:
@jondeandres :facepalm: thanks for pointing out those obvious errors -- I was tired last night when I thew that together.

I've now fixed everything and also split it up into multiple commits.

There are no tests for this, perhaps we can write some before merging.

jjb:
Are there tests for something else that I can look at to base tests for this off of?

jondeandres:
@jjb cool :+1: 

you can write a test for this using the inline option in sidekiq, https://github.com/mperham/sidekiq/wiki/Testing. Just write a job that will crash, enqueue it, and check that the report was sent. `Rollbar.last_report` should return something after the report is sent.

I'll be on freenode as 'jandres' if you want to chat it.

jjb:
@jondeandres thanks

seems like the block given to `Sidekiq.configure_server` does not get executed in the testing environment

i'm exploring why that might be... let me know if you have an idea

jjb:
@jondeandres okay I figured out it's because of my sidekiq test environment -- I submitted this ticket, we'll see what @mperham says https://github.com/mperham/sidekiq/issues/2128

jondeandres:
it could because this?

https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L53
https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L67

require 'sidekiq/cli' and it should work.

But now I'm thinking that perhaps middlewares will not be executed in tests environments? I've never tested Sidekiq jobs, but probably Sidekiq will have this documented.

jjb:
@jondeandres finally got around to finishing this up, take a look

jjb:
when #209 is merged i'll rebase, for a tidier history on this one

jondeandres:
Thanks @jjb. Tests are failing cause sidekiq is not supported in Ruby 1.8.7, take a look at our gemspec file, https://github.com/rollbar/rollbar-gem/blob/master/rollbar.gemspec#L25.

jjb:
@jondeandres any idea why that lone failure happened? https://travis-ci.org/rollbar/rollbar-gem/jobs/49406361

```
/home/travis/.rvm/gems/ruby-2.1.0/gems/genspec-0.2.7/lib/genspec.rb:8:in `<top (required)>': Unsupported Rails version: 4.0.4 (RuntimeError)
	from /home/travis/.rvm/gems/ruby-2.1.0/gems/activesupport-4.0.4/lib/active_support/dependencies.rb:229:in `require'
	from /home/travis/.rvm/gems/ruby-2.1.0/gems/activesupport-4.0.4/lib/active_support/dependencies.rb:229:in `block in require'
```

jondeandres:
@jjb is because the version of genspec:

0.2.7 https://github.com/sinisterchipmunk/genspec/blob/v0.2.7/lib/genspec.rb
0.2.8 https://github.com/sinisterchipmunk/genspec/blob/v0.2.8/lib/genspec.rb

It seems to not be crashing in master, so probably is something in your branch. You can change gemfiles/rails40.gemfile and check that is solved:

```
gem 'genspec', '> 0.2.7'
```

jjb:
@jondeandres gotcha. i increased the version for the whole project, no reason to not do this right?

(after it runs i'll check it to only be for rails 4 if you prefer that)

jondeandres:
Nice!

One comment in specs. Can you replace this style in general?

`let(:foo){:bar}` to `let(:foo) { :bar }`

jjb:
@jondeandres all set

spec/rollbar/sidekiq_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'spec_helper'
+
+unless RUBY_VERSION == '1.8.7'
+  require 'sidekiq'
+  require 'rollbar/sidekiq'
+end
+
+describe "Sidekiq integration", :reconfigure_notifier => false do
```
jondeandres: This should be `Rollbar::Sidekiq`.

spec/rollbar/sidekiq_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  before{ skip if RUBY_VERSION == '1.8.7' }
+
+  describe "Rollbar::Sidekiq.handle_exception" do
```
jondeandres: this `describe '.handle_exception'`

spec/rollbar/sidekiq_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    let(:msg_or_context) { ["hello", "error_backtrace", "backtrace", "goodbye"] }
+    let(:exception) { StandardError.new("oh noes") }
+    let(:rollbar) { double }
+    subject do
+      Rollbar::Sidekiq.handle_exception(msg_or_context, exception)
```
jondeandres: I prefer to see the 'trigger' statement in the `it` block. And subject should be related to the object you are testing. In this case `described_class` will return `Rollbar::Sidekiq`.

You can change this to `subject { described_class }`. And use later `subject.handle_exception(msg_or_context, exception)` in the `it` block.

spec/rollbar/sidekiq_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    it "constructs scope from filtered params" do
+      rollbar.stub(:error)
+      Rollbar.should_receive(:scope).with(
```
jondeandres: Can we use new rspec syntax?

`allow(Rollbar).to receive(:scope).with(....)`

spec/rollbar/sidekiq_spec.rb
jjb: Then it will blow up under 1.8.7 because that class isn't loaded in lines 3-6 above

spec/rollbar/sidekiq_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+    it "sends the passed-in error to rollbar" do
+      Rollbar.stub(:scope) { rollbar }
+      rollbar.should_receive(:error).with(exception, :use_exception_level_filters => true)
+      subject
+    end
+  end
+
+  describe "middleware" do
+    let(:middleware) { Rollbar::Sidekiq.new }
```
jondeandres: This is really the `subject`, and we are describing `#call`.

You can have:

```ruby
let(:middleware_block) { proc { raise exception } }
subject { described_class.new }

it '...' do 
  expect { subject.call(nil, msg, nil, &middleware_block) }.to raise_error(exception)
```

spec/rollbar/sidekiq_spec.rb
jondeandres: Humm! And can we use an string here then instead of a constant?

jjb:
@jondeandres perhaps it will be more efficient for you to simply fork my branch and make the style changes you want? It's probably faster to make the changes directly than to explain to me the rationale.

jondeandres:
@jjb yah! probably! I'll for this and make a PR to your repo. I hope we can merge this today. Sorry for this style comments, it's just to maintain some order here :-D.

jjb:
no problem, i enjoyed the process.

jjb:
@jondeandres merged

jondeandres:
Nice! Thanks @jjb for this!

jjb:
@jondeandres were you going to merge this one today?

jondeandres:
Merged, thanks @jjb!!

jjb:
:eggplant:

PR #198, opened 2015-01-08
Dump payload when we are writing to file.

jondeandres:


lib/rollbar.rb
```diff
@@ -517,7 +517,7 @@ def do_write_payload(payload)
           @file = File.open(configuration.filepath, "a")
         end
 
-        @file.puts payload
+        @file.puts(dump_payload(payload))
```
brianr: I kinda think that we should do this first, before opening the file, to make things simpler in case the dump fails. Thoughts?

jondeandres: You are right @brianr, I've just changed it.

jondeandres:
@brianr tests are passing here, I think we can merge this and release a new version.

brianr:
Released as 1.3.2

PR #199, opened 2015-01-13
Use :use_exception_level_filters option in middlewares.

jondeandres:
Now by default we will not override the level using exception_level_filters if :use_exception_level_filters is not true.

### DIFF
--- a/spec/rollbar/middleware/sinatra_spec.rb
+++ b/spec/rollbar/middleware/sinatra_spec.rb
@@ -42,7 +42,7 @@ describe Rollbar::Middleware::Sinatra, :reconfigure_notifier => true do
   end
 
   let(:expected_report_args) do
-    [uncaught_level, exception]
+    [uncaught_level, exception, { :use_exception_level_filters => true}]
   end
 
   describe '#call' do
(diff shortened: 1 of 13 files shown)


jondeandres:
@brianr what do you think we should do with this https://github.com/rollbar/rollbar-gem/blob/use_exception_level_filters-option/lib/rollbar.rb#L128 ?

brianr:
@jondeandres 'ignored' should follow the same rules as setting a level. That is, it should now only apply when :use_exception_level_filters is true.

jondeandres:
@brianr ok!

jondeandres:
@brianr https://github.com/rollbar/rollbar-gem/commit/8e656451bee557db0c96a36f34fa36b98941c433

spec/rollbar/middleware/sinatra_spec.rb
```diff
@@ -42,7 +42,7 @@ def app
   end
 
   let(:expected_report_args) do
-    [uncaught_level, exception]
+    [uncaught_level, exception, { :use_exception_level_filters => true}]
```
brianr: weird spacing here, did you mean to do it like that?

jondeandres: not, it should be `{ :use_exception_level_filters => true }` as most of ruby style guides recommend.

I'll fix it.

jondeandres: Fixed

brianr:
Looks good, and verified that this works as expected. Will push out a new release shortly.

PR #202, opened 2015-01-16
Add person data docs and rack examples in README.md.

jondeandres:


jondeandres:
@brianr can we merge this?

PR #203, opened 2015-01-16
Allow to send a custom message on the extra hash

olvap:
I didn't find any way to send a custom message in the implementation, so I added this line to send it on the extra hash.
You will have to use it like this

~~~ ruby
Rollbar.info "title", custom: {some: 'value'}
~~~

I didn't created any test because i not sure if it is the correct approach. What do you think?

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -293,6 +293,7 @@ module Rollbar
         }
       }
 
+      data[:custom] = extra.delete(:custom) if extra && extra[:custom]
       data[:body] = build_payload_body(message, exception, extra)
       data[:project_package_paths] = configuration.project_gem_paths if configuration.project_gem_paths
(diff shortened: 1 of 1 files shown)


brianr:
Thanks @olvap !

The intended way to do it now is like this:

```ruby
Rollbar.scope({:custom => {:some => "value"}}).info("title")
```

Does that make sense?

We should document this better, and maybe add a helper method...

olvap:
In that case, I could added to the doc.

olvap:
May be a helper is a good idea.

jondeandres:
@brianr what's the idea about this? if we decide to set the `custom` value using `scope` or `scoped` then we should document it. 

@olvap having an extra argument that takes part for the `extra` value and part for `custom` value can be confusing IMHO. But it's clear we should improve our docs or the interface for this.

olvap:
I'll close this PR for now.

PR #206, opened 2015-01-20
Add backtrace to exceptions without backtrace. Closes #205.

### LINKED ISSUE #205: Add stacktrace when reporting exception manually
I wondered why I never saw a backtrace in Rollbar when I reported an error with `Rollbar.error MyError.new`. I mailed your support and @jondeandres explained that errors created with `.new` do not have a backtrace. (Thanks, Jon!)
However, it's quite easy to add:

```ruby
e = Exception.new
e.set_backtrace caller
```

What do you guys think about adding a backtrace to any error that is being reported automatically if it doesn't have one yet?

jondeandres:


lib/rollbar.rb
```diff
@@ -363,6 +360,18 @@ def trace_data(exception)
       }
     end
 
+    def exception_backtrace(exception)
+      return exception.backtrace if exception.backtrace.respond_to?( :map )
+      return [] unless configuration.populate_empty_backtraces
+
+      cleaned_backtrace = caller.reject {|path| path =~ /#{rollbar_gem_dir}/ }
```
brianr: Does this reject all rollbar frames anywhere in the stack? I think we should keep the ones that appear in the middle of the stack (i.e. our middleware).

jondeandres: Good point @brianr.

That's the issue, if we don't clean that filenames we will always show frames from the `Rollbar.error()` statement. But clean only those frames can be a little tricky.

Would you like to clean it in a smarter way? Not clean? Don't provide this feature?

lib/rollbar.rb
```diff
@@ -363,6 +360,19 @@ def trace_data(exception)
       }
     end
 
+    def exception_backtrace(exception)
```
brianr: Nice. Some comments here would be helpful.

jondeandres:
@brianr added comments and moved ruby-debug and byebug to Gemfile.

PR #208, opened 2015-01-23
Parse configuration.payload_options when reporting internal errors.

jondeandres:
This avoids sending null values for 'context' key in data. Our API only
supports string values for context.


jondeandres:
Pass already passing, this can be merged.

PR #210, opened 2015-02-03
Defaults shouldn't override previously set variables

kdonovan:
In our application we set the rollbar_token from environmental variables for security, but when rollbar loads itself up it blows away the already-set variables with it's defaults.

### DIFF
--- a/lib/rollbar/tasks/rollbar.cap
+++ b/lib/rollbar/tasks/rollbar.cap
@@ -34,9 +34,9 @@ end
 
 namespace :load do
   task :defaults do
-    set :rollbar_user,  Proc.new { ENV['USER'] || ENV['USERNAME'] }
-    set :rollbar_env,   Proc.new { fetch :rails_env, 'production' }
-    set :rollbar_token, Proc.new { abort "Please specify the Rollbar access token, set :rollbar_token, 'your token'" }
-    set :rollbar_role,  Proc.new { :app }
+    set :rollbar_user,  fetch(:rollbar_user,  Proc.new { ENV['USER'] || ENV['USERNAME'] } )
+    set :rollbar_env,   fetch(:rollbar_env,   Proc.new { fetch :rails_env, 'production' } )
+    set :rollbar_token, fetch(:rollbar_token, Proc.new { abort "Please specify the Rollbar access token, set :rollbar_token, 'your token'" } )
+    set :rollbar_role,  fetch(:rollbar_role,  Proc.new { :app } )
   end
 end


brianr:
Thanks @kdonovan! This looks good to me. @jondeandres can you review as well?

jondeandres:
@brianr it's nice for me.

brianr:
Merge, thanks @kdonovan! Will get this out in the next release.

brianr:
Released in 1.4.3 (now on rubygems)

shaneog:
This is breaking deploys for me. The problem is that by the time the Rollbar `load:defaults` task is run, the `deploy.rb` file has not been parsed, and thus my `set :rollbar_token, ENV['ROLLBAR_ACCESS_TOKEN']` has not been run.
So I'm consistently getting the message `Please specify the Rollbar access token, set :rollbar_token, 'your token'`


@kdonovan How are you loading the Roller access token ENV variable?

jondeandres:
I've detected the problem here.

`fetch` is not so safe cause it evaluates the default value if it's callable, https://github.com/capistrano/capistrano/blob/master/lib/capistrano/configuration.rb#L41.

We can do something crazy like `Proc.new { Proc.new { ... } }`, but is really ugly.

@kdonovan could you provide us the loading order you have when setting `:rollbar_token`, `require 'rollbar/capistrano3` and others?

It seems you are setting `:rollbar_token` before `require 'rollbar/capistrano3' is executed, could be?

My suggestion is change that order, and we'll then revert this PR since it's being a problem for most of the users.

jondeandres:
@kdonovan I could reproduce your problem with this code

```ruby
task :foo do
  set :rollbar_token, proc { 'foo' }
end

before 'load:defaults', 'foo'
```
Are you using something similar? In that case 'load:defaults' will override your values. 

We can fix this for everybody with this code:

```ruby
namespace :load do
  task :defaults do
    set :rollbar_user,  Proc.new { ENV['USER'] || ENV['USERNAME'] } unless fetch(:rollbar_user)
    set :rollbar_env,   Proc.new { fetch :rails_env, 'production' } unless fetch(:rollbar_env)
    set :rollbar_token, Proc.new { abort "Please specify the Rollbar access token, set :rollbar_token, 'your token'" } unless fetch(:rollbar_token)
    set :rollbar_role,  Proc.new { :app } unless fetch(:rollbar_role)
  end
end
```

But probably the best is just not set anything before 'load:defaults' is executed, https://github.com/capistrano/capistrano/blob/master/lib/capistrano/setup.rb#L13.

PR #212, opened 2015-02-04
Add convenience scope! method to Rollbar public interface.

jondeandres:


brianr:
Works great. Let's make sure the tests pass, and then we can push out a new release.

PR #217, opened 2015-02-11
Support Lotus Framework [WIP]

joneslee85:
WIP


jondeandres:
Hey!

Thank you for this Pull Request, it's always valuable have active customers  :-D. The code itself looks good, but I think it's not needed and this code is responsibility of the gem user.

I mean, although Lotus framework could be used by many people, this code doesn't provide any new feature in spite the automatic configuration where it'll be always needed use `require 'rollbar/lotus'`.

You are adding about 8 new builds to our CI without really have added any new spec.

It's nice, but I think that this just resolve the configuration for some specific projects, and it's really acceptable to be configured with an initializer, just requiring the rollbar configuration file.

It'd be nice however having more integration with Lotus framework and more specific with [Lotus Controller](https://github.com/lotus/controller). We'll take a look into this.

Thank you for the PR!

brianr:
Hey @joneslee85  -- thanks for the PR! We'd be happy to support Lotus. Let us know if you run into any problems, and when it's ready for us to take a deeper look.

joneslee85:
@jondeandres Understood. This is WIP and I am struggling to get lotusrb gem introduced as development dependency for its conflicts with rails 3.0 gem. I am going to submit a separate PR so remove rails 3 from gemspec development dependency.

@brianr Much thanks for your support. I'll let you know once it's ready for review.

joneslee85:
@brianr what do you think about grouping specs by frameworks? I am planning to remove all details about development dependency from gemspec and moving it to Appraisal instead, the reason behind that is because lotusrb conflicts with rails.

jondeandres:
@joneslee85 for me sounds good, although can be a little tricky the daily development here, but probably is the best. show us the commit with the changes you are proposing and let's talk about it :-D.

Thanks @joneslee85!!

joneslee85:
@jondeandres https://github.com/rollbar/rollbar-gem/pull/222

PR #219, opened 2015-02-11
Test against Rails 4.2.0

joneslee85:
ready to rock


spec/generators/rollbar/rollbar_generator_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe :rollbar do
+describe 'generator', :rollbar do
```
jondeandres: @joneslee85 tests are failing cause this, https://github.com/rails/rails/blob/v3.0.20/railties/lib/rails/generators.rb#L128, cannot load the generator.

genspec can however load the generator using the class constant. If you change this line with the next specs should pass:

```ruby
require 'generators/rollbar/rollbar_generator'

describe Rollbar::Generators::RollbarGenerator do
```

jondeandres:
@joneslee85 thank you for this!

Just check the comment I've done about genspec and generators. I think with that change this can be ready to be merged.

joneslee85:
@jondeandres as I've suggested before, what do you think about having separate dummy app for each rails version? I find there are so many if..else condition as more and more Rails versions get released.

jondeandres:
@joneslee85 I'd prefer having a smarter dummy app that loads different configuration files depending on the rails version.

The if/else statements you are commenting are only in config files and actually in `app/models/user.rb`. What if just we create a module to load some files depending on the Rails version?

While the controllers implementation works for all the rails version I'd prefer not having different applications. If in the future this cannot be avoided then we can change it.

brianr:
Hey @joneslee85 , did you mean to close this PR? Jon and I were just discussing and I think it could be merged - let me know what you think.

PR #222, opened 2015-02-20
Group spec by framework

joneslee85:
Restructure the spec by grouping specs by framework:

* spec/rails     -> rails-related
* spec/rollbar  -> non-rails related
* spec/sinatra  -> sinatra-related


joneslee85:
@jondeandres @brianr I am scratching my head with this failed test:

`bundle exec rspec ./spec/rails/requests/home_spec.rb:39`

I am sure ControllerMethods aren't loaded correctly. Wondering if any of you could pair with me on this. Cheers

jondeandres:
@joneslee85 this looks pretty good! I'll fetch your branch and look at that spec that is failing.

thank you for this, it's awesome!

joneslee85:
@jondeandres any idea?

jondeandres:
no yet, let me today to take a look into this.

thanks again @joneslee85!

joneslee85:
ping

jondeandres:
@joneslee85 couldn't focus on this yet, sorry.

btw, there are two kind of errors for the same test, did you realize?

one question @joneslee85, could you explain us exactly wich incompatibilities exist between Lotus and Rails? And why cannot be both loaded?

The code looks good, it's a good job.The problem with this change is that the future development will not be so fluent as now, where just `bundle exec rspec` will give you the status of the specs. Isn't any other solution for this?

jondeandres:
@joneslee85 you'll need to update `travis.yml` in order to build the specs for the new gemfiles to have.

jondeandres:
@joneslee85 

here is the problem:

```diff
diff --git a/spec/rails/requests/home_spec.rb b/spec/rails/requests/home_spec.rb
index 7f03016..35efce5 100644
--- a/spec/rails/requests/home_spec.rb
+++ b/spec/rails/requests/home_spec.rb
@@ -3 +3 @@ require 'rails_helper'
-describe HomeController, :type => :controller do
+describe HomeController, :type => :request do
```

joneslee85:
@jondeandres Lotus has conflict with tilt version that is shipped along Rails 3.2. As Lotus grows, I would see more conflicts. As much as I do not want to separate specs, there are no proper way to get 2 of them playing nice together under same Gemfile roof.

Regarding the workflow of running spec, I am thinking of a quick script that would help you set up and test against all frameworks in 1 go. I'll see what I could come up with.

Thanks much for your help.

joneslee85:
@jondeandres on the second look today, what do you think about a bash script that loop through the gemfiles then set BUNDLE_GEMFILE and run rake spec?

jondeandres:
@joneslee85 that can be an option, the only problem is that we'll have to run `bundle` before every `rake spec`. BTW, can web run a single spec with `rake spec`? Looking at the task I'd say not.

joneslee85:
@jondeandres I will have a look at this today, hopefully I could come up with something sensible.

joneslee85:
@jondeandres what do you think about extracting the core part to a gem so users could easily provide integration gem?

joneslee85:
I am thinking about having a gem called `rollbar-core` and `rollbar-rails` (which is for Rails) OR `rollbar`

jondeandres:
Hi @joneslee85!!

 I guess your problem is https://github.com/rollbar/rollbar-gem/blob/master/rollbar.gemspec#L21-L22 no?

Do you find any way to resolve this, having your own `rollbar-lotus` gem, without move all the Rails stuff to a new gem? 

The main problems I find here is:
* it'll become our code more distributed and become more complex to maintain. This is not a problem if this really resolves something. Your problem could be a reason yes.
* Most important. The customers may have rollbar installed with something like `gem 'rollbar'`, they will upgrade and nothing will work for them, since they need `rollbar-rails`.

Let me discuss this with my colleagues ok? :smile:

joneslee85:
@jondeandres ping :)

jondeandres:
@joneslee85 sorry for the delay.

we will probably need that `rollbar` gem depends on `rollbar-core` and `rollbar-rails` so we don't break anything on customer's implementations. Or having all the Rails stuff in `rollbar`. The way to use Lotus then is adding `rollbar-lotus` to customer's Gemfile.

Have sense for you?

Thanks @joneslee85 :smile:

joneslee85:
@jondeandres Yes it does. I'd go for the first option.

joneslee85:
@jondeandres what's the plan? Do you need any help from me?

jondeandres:
@joneslee85 sorry for the delay. mmmm, well, Ithink we could split rollbar gem into rollbar-core and rollbar-rails and have rollbar gem depending on both. I cannot say you when we can start doing this, so you are welcome to open a PR with the basic idea and can work on it together.

BTW, I think I'll take a look in depth at Lotus, seems to be promising :smile:

brianr:
@jondeandres @joneslee85 What do you think of the following structure:

- `rollbar-core`
- `rollbar` - depends on `rollbar-core`, and contains all of the same hooks that it does today, including rails, etc.
- `rollbar-lotus` - depends on `rollbar-core`; contains only the hooks needed for lotus

In the future we can split out the other parts of the `rollbar` gem into `rollbar-rails`, `rollbar-delayed-job`, `rollbar-sidekiq`, and so on. But this way we don't break anything for the users who currently have `rollbar` installed.

jondeandres:
@brianr it sounds good for me. it's a good first step.

joneslee85:
Sound a geat idea! Can you please help create new repos?

jondeandres:
@joneslee85 I've created the repo https://github.com/rollbar/rollbar-ruby-core. You can fork it and open PRs to our repo. We can then discuss over the PRs the implementation.

Thanks!

joneslee85:
:+1:

PR #223, opened 2015-02-25
Move `ActiveRecord` patching into `require_hooks`

brandur:
This allows it to respect the `disable_monkey_patch` configuration option like the extensions for Rack, Rake, Sidekiq, etc. already do.

We have a project that still uses an ActiveRecord 2* version and this patch doesn't apply itself correctly to versions from that era. More generally, we're also just trying to reduce the amount of monkey patching that we're doing overall.

Are there any known issues with activating the `ActiveRecord` patch at the same time as the others? Thanks!

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -14,7 +14,6 @@ require 'rollbar/version'
 require 'rollbar/configuration'
 require 'rollbar/logger_proxy'
 require 'rollbar/exception_reporter'
-require 'rollbar/active_record_extension' if defined?(ActiveRecord)
 require 'rollbar/util'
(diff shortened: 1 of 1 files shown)


jondeandres:
@brandur thanks for the PR. it make sense moving that require to `.require_hooks` method. let's wait for the specs to pass and we can merge it.

thanks!!

brandur:
@jondeandres Thanks for the crazy quick review! And I'll say that this is one seriously comprehensive build process! Hah.

brandur:
Thanks again Jon!

jondeandres:
Thanks to you ;-)

PR #228, opened 2015-03-18
Use appraisal 1.0.2 for rails 3.0 and rails 3.1. Those are the used by ruby 1.8.7

jondeandres:


jondeandres:
Let's wait for the tests.

PR #229, opened 2015-03-19
Config Template

revans:
Set "config.enabled" to false for the development environment as well as the test environment.

### DIFF
--- a/lib/generators/rollbar/templates/initializer.rb
+++ b/lib/generators/rollbar/templates/initializer.rb
@@ -12,8 +12,8 @@ Rollbar.configure do |config|
   <% else %>
   config.access_token = <%= access_token_expr %>
 
-  # Here we'll disable in 'test':
-  if Rails.env.test?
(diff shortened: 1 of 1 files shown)


brianr:
@revans, appreciate the PR! We want to keep 'development' enabled by default, though. The main benefits we see are:

- Rollbar code gets executed in development, so you're more likely to notice if any problems arise. This is especially relevant if you start adding Rollbar.error() calls throughout your application.
- I personally really like the "View in Rollbar" link (in the logs) - it makes it really easy to share an exception with a co-worker. I'll frequently post links in our Slack room to the Rollbar page about an error I just ran into in my dev instance
- In the future, we'll help you find similar issues that may have happened in other environments (i.e. you hit some bug in development, and then later something similar appears in production). We do this a little bit now on the "Similar Items" tab.
- In the interface, you can select specific environment (i.e. "Production") so clutter hasn't been an issue for us. And of course, you're always welcome to disable it in your own setup.

PR #245, opened 2015-05-25
Fix error in code example for custom Async handlers

radanskoric:
There is an error in code example for adding a custom async handler that will actually not activate async handling at all. 

Also, my sublime config removed some blank spaces at ends of lines, hope that is ok.


jondeandres:
Hey @radanskoric, thank you for this!!

radanskoric:
np, glad to help :)

PR #250, opened 2015-06-11
Add Encoding module, with Encoder and LegacyEncoder classes.

jondeandres:


lib/rollbar/encoding/encoder.rb
```diff
@@ -0,0 +1,54 @@
...
+      def initialize(object)
+        @object = object
+      end
+
+      def encode
+        value = object.to_s
+
+        encoded_value = force_encoding(value).encode(*encoding_args(value))
+
+        object.is_a?(Symbol) ? encoded_value.to_sym : encoded_value
```
coryvirok: This may be a dumb question, but what is this line for? Why do we need to turn the string back into a symbol?

jondeandres: mmmm, yah. We don't need that. I'll change it.

lib/rollbar/encoding.rb
```diff
@@ -0,0 +1,22 @@
...
+    def self.encode(object)
+      can_be_encoded = object.is_a?(Symbol) || object.is_a?(String)
+
+      return object unless can_be_encoded
+
+      encoding_class.new(object).encode
+    end
+
+    def self.encoding_class
+      if String.instance_methods.include?(:encode)
```
joneslee85: What do you think about moving `require` under this conditional block?

lib/rollbar/encoding/encoder.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def detect_encoding(v)
+        value = v.dup
+        ALL_ENCODINGS.detect do |encoding|
+          begin
+            value.force_encoding(encoding).encode(::Encoding::UTF_8).codepoints
+          rescue
+            false
+          else
+            true
```
joneslee85: what do you think about moving this line behind line 36?

spec/rollbar/encoding/encoder_spec.rb
```diff
@@ -0,0 +1,63 @@
+# encoding: utf-8
+require 'spec_helper'
+
+unless RUBY_VERSION.start_with?('1.8')
+  require 'rollbar/encoding/encoder' unless
```
joneslee85: I find the `unless` is a bit confusing.

Btw, I am wondering if you should unit test this Encoder class? What do you think about treating this class as private and test the Encoding class instead?

lib/rollbar/encoding.rb
jondeandres: I like that, thanks! I'll change it.

lib/rollbar/encoding/encoder.rb
jondeandres: Probably better, thanks.

spec/rollbar/encoding/encoder_spec.rb
jondeandres: That `require` and `unless` is something weird. I'll review it.

@joneslee85 I understand your point, but I'd prefer to just test that class separately.

PR #254, opened 2015-07-08
Remove MultiJson and use just native or gem JSON.

jondeandres:
We are having some problems serializing the payload under some scenarios. One scenario very common is when customers are using https://github.com/roidrage/redis-session-store.

We are adding to the payload `env['rack.session.options']` from Rack env. When using Redis as session store that value will have a `Redis::Client` instance.

Beside this, when ActiveSupport `Object` JSON extensions are loaded, serializing the payload cause an infinite loop in most of cases. This can vary depending on the serializing library the customer has installed in his project.

I cannot achieve the exact reasons of this, but I think it's a problem serializing `TCPSocket`. I've changed https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/json.rb adding this:

```ruby
class Socket
  def as_json
    to_s
  end
end
```

And it worked correctly, providing a huge session options `Hash`. this is not exacty what we want. And at the same time, the serializing was quite slow.

The problem is exploited since `MultiJson` will call other libraries, and those libraries will call `object.to_json` most of the cases.

In Oj they have taken account of that in some of their functions, https://github.com/ohler55/oj/blob/master/ext/oj/oj.c#L1819.

But MultiJson has `to_json: true` by default in Oj adapter, https://github.com/intridea/multi_json/blob/master/lib/multi_json/adapters/oj.rb#L9.

I've decided to just load JSON library and use it, seems to be the best solution in many other libraries I've reviewed. JSON will be wrapped by `Rollbar::JSON`, that loads it and calls `JSON.generate` and `JSON.load`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 sudo: false
+services:
+  - redis-server
 language: ruby
 # Broken bundler on travis CI - https://github.com/bundler/bundler/issues/2784
 before_install:
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -11,6 +11,7 @@ rescue LoadError
 end
 
 require 'rollbar/version'
+require 'rollbar/json'
 require 'rollbar/configuration'
 require 'rollbar/logger_proxy'
 require 'rollbar/exception_reporter'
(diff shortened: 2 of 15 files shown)


brianr:
Nice work @jondeandres!

Could you add some tests for things that used to crash (like `TCPSocket`) to ensure that they don't crash now?

jondeandres:
@brianr I'll see if I can reproduce this on a test. Maybe it'll not be easy.

PR #259, opened 2015-07-15
Filter uploaded files as part of array param

mdominiak:
When request parameter is an array of `ActionDispatch::Http::UploadedFile` then uploaded files are not filtered in `rollbar_filtered_params` which in turn can cause `Failsafe from rollbar-gem: error in process_payload` when uploading binary files.

Example request parameters for `<input type="file" multiple name="files[]" />` used with jquery-fileupload:

```
Parameters: {"utf8"=>"✓", "authenticity_token"=>"[FILTERED]", "files"=>[#<ActionDispatch::Http::UploadedFile:0x007fec8f110fe8 @tempfile=#<Tempfile:/var/folders/bp/mg0pdsts6x17l3ft6h2dbj000000gn/T/RackMultipart20150715-8767-kojuih.png>, @original_filename="file.png", @content_type="image/png", @headers="Content-Disposition: form-data; name=\"files[]\"; filename=\"file.png\"\r\nContent-Type: image/png\r\n">]}
```

### DIFF
--- a/lib/rollbar/request_data_extractor.rb
+++ b/lib/rollbar/request_data_extractor.rb
@@ -166,19 +166,25 @@ module Rollbar
           result[key] = rollbar_filtered_params(sensitive_params, value)
         elsif value.is_a?(Array)
           result[key] = value.map do |v|
-            v.is_a?(Hash) ? rollbar_filtered_params(sensitive_params, v) : v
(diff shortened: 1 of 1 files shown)


jondeandres:
@mdominiak it looks good to me. It'd be awesome have a test in `spec/controllers/home_controller_spec.rb` for this.

If you cannot do it we'll take this and do the test in next days.

Thanks for this!

PR #261, opened 2015-07-17
Monkey patch BasicSocket just for active_support < 4.1.0.

jondeandres:


brianr:
:+1:

lib/rollbar.rb
```diff
@@ -736,8 +736,19 @@ def require_core_extensions
...
       return if configuration.disable_core_monkey_patch
 
-      # Needed to avoid active_support bug serializing JSONs.
-      require 'rollbar/core_ext/basic_socket'
+      # Needed to avoid active_support (< 4.1.0) bug serializing JSONs
+      require 'rollbar/core_ext/basic_socket' if monkey_patch_socket?
+    end
+
+    def monkey_patch_socket?
+      return false unless defined(ActiveSupport::VERSION::STRING)
```
joneslee85: what do you think about using `defined?`?

lib/rollbar.rb
```diff
@@ -736,8 +736,19 @@ def require_core_extensions
...
+
+      major, minor = ActiveSupport::VERSION::STRING.split('.').map(&:to_i)
+
+      return true if major == 3
+      return true if major == 4 && minor == 0
```
joneslee85: what do you think about consolidating 748 and 749 into 1 line?

lib/rollbar.rb
```diff
@@ -736,8 +736,19 @@ def require_core_extensions
 
       return if configuration.disable_core_monkey_patch
 
-      # Needed to avoid active_support bug serializing JSONs.
-      require 'rollbar/core_ext/basic_socket'
+      # Needed to avoid active_support (< 4.1.0) bug serializing JSONs
+      require 'rollbar/core_ext/basic_socket' if monkey_patch_socket?
```
joneslee85: What do you think about moving this under `rollbar/rails` directory?

lib/rollbar.rb
jondeandres: Oh yes, This should be `defined?` yes. Thanks!

lib/rollbar.rb
jondeandres: For me it's pretty clear in this way, thanks for the tip.

lib/rollbar.rb
jondeandres: While we don't refactor the gem, I'd prefer this code in this file. Thanks.

Issue #263, opened 2015-07-20
Recommended ActiveJob integration


monkbroc:
Since ActiveJob has been part of Rails for a while now it would be great if the README included a recommended integration between Rollbar and ActiveJob so that any uncaught exception in a background job is reported.

I'm using SuckerPunch as the backend and I thought that `config.use_sucker_punch` would report the errors, but it only queues a job through sucker punch to report already caught errors.

Here's my integration. Any better ideas?

```ruby
class BaseJob < ActiveJob::Base
  # Report any errors in jobs to Rollbar
  rescue_from Exception do |exception|
    Rollbar.error(exception, job: self.class.name, job_id: job_id)
  end
end
```

jondeandres:
@monkbroc that snippet should be enough, yes. Maybe we should comment it in our README.

We could provide a module that the users will include in their jobs, but I'm not sure if this will be more complex for them or it's better just let them send the errors manually.

@monkbroc do you have any preference about this?

Would you like this?

```ruby
class YourAwesomeJob < ActiveJob::Base
  include Rollbar::ActiveJob
end
```

monkbroc:
A module with a working default would be great.

I made a PR #264 for integrating with ActiveJob.

DMA57361:
I've just tried this out - to replace some custom error handling we had that did a similar job - and thought I'd point out that this can cause some unexpected behaviour because it swallows unhandled exceptions.

I noticed this because we use Sidekiq as our ActiveJob backend and it will automatically retry failed jobs. However, this stopped happening when I activated this because the exceptions were caught and the job was instead considered to have been successful.

Since we've only the one project using ActiveJob at the moment I don't know what effect, if any, this could have on most of the other backends. It also seems to cause jobs called with `perform_now` or via the inline backend to fail for unhandled exceptions (which they would otherwise - the exception is just allowed to bubble up),

I think either a reraise or maybe warning in the readme would be useful.

jondeandres:
Hi @DMA57361, thanks for the tip.

I think this kind of behaviors will be a problem between the ActiveJob API and the backends. I'm not sure about which solution is the best. I'll investigate a bit about this.

The philosophical problem I find is if, in your case, Sidekiq should determine if it should retry or not, when ActiveJob has its own API to define that.

And I'm not sure about the reraise inside `rescue_from`. I'll test it. Maybe that's the best solution yes.

DMA57361:
[...]

PR #270, opened 2015-08-03
Reraise the error so ActiveJob backend can retry

monkbroc:
When using ActiveJob, some backends have a default behavior when rescueing an error like requeueing the job and retrying. Reporting the error to Rollbar should not change this behavior.


monkbroc:
This addresses a concern by @DMA57361 in #263 that swallowing the errors in ActiveJob is a bad idea.

jondeandres:
Thanks @monkbroc!

brianr:
Thanks @monkbroc!

Build is passing now - a few parts of the matrix failed before because some gems couldn't be downloaded, but succeeded when retried.

PR #276, opened 2015-08-14
allows access_token override in extras

ptrln:
This allows an access_token to be provided when logging. This is useful to put rollbars into the correct project.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -301,6 +301,9 @@ module Rollbar
         }
       }
 
+      access_token = configuration.access_token
+      access_token = extra.delete(:access_token) if extra && extra[:access_token].present?
+
       data[:body] = build_payload_body(message, exception, extra)
       data[:project_package_paths] = configuration.project_gem_paths if configuration.project_gem_paths
       data[:code_version] = configuration.code_version if configuration.code_version
@@ -317,7 +320,7 @@ module Rollbar
       data.delete(:context) unless data[:context]
 
       payload = {
-        'access_token' => configuration.access_token,
+        'access_token' => access_token,
         'data' => data
       }
 
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -302,6 +302,10 @@ describe Rollbar do
           payload.keys.should match_array(['access_token', 'data'])
         end
 
+        it 'should have the correct access_token' do
+          payload['access_token'].should == Rollbar.configuration.access_token
+        end
+
         it 'should have the correct data keys' do
           payload['data'].keys.should include(:timestamp, :environment, :level, :language, :framework, :server,
             :notifier, :body)
@@ -325,6 +329,14 @@ describe Rollbar do
         it 'should have the correct level and message body' do
           payload['data'][:level].should == 'info'
           payload['data'][:body][:message][:body].should == 'message'
+        end
+
+        context 'with access_token passed in' do
+          let(:extra_data) { {:key => 'value', :hash => {:inner_key => 'inner_value'}, :access_token => 'my-other-access-token'} }
+
+          it 'should use the access_token passed in' do
+            payload['access_token'].should == 'my-other-access-token'
+          end
         end
       end
 


jondeandres:
Hi @ptrln, thanks for the PR.

We'll merge it and add some docs in the README.md

meetrajesh:
Damn, I've really wanted this feature for a long time.

+1

jondeandres:
Hi @ptrln and @meetrajesh,

after think a bit more about this implementation, we think we are going to revert it. With this we are breaking that somebody could just report a `access_token` value to Rollbar, just as extra data.

We'll try to do another approach with this, probably using Rollbar.scoped().

brianr:
@ptrln, thanks for your work on this!

Elaborating on what @jondeandres said, the trouble with putting it in `extra` is that it conflicts with actual data where the param name is `access_token`--classic in-band signaling. This change would cause any existing code that does that to suddenly break, and any new code to have surprising behavior (you'd need to know that the extra params are just data, except if the param name is 'access_token'...).

The alternatives I can think of are:

- `Rollbar.scope(:access_token => "new token")` - pretty close, but potential in-band signaling there too. Everything you pass to `Rollbar.scope` is supposed to go directly inside of the `data` part of the final payload; if you try this in the current release, you will have an `access_token` param as a sibling to `request`, `body`, etc.
- `Rollbar.scope(nil, {:access_token => "new token"})` - i.e. make the second arg affect the config options that aren't inside the 'data' part of the payload, like 'access_token', scrubbing options, exception_level_filters, and so on.
- `Rollbar.scope_config({:access_token => "new token"}) - same as previous, but with a new method call

At the moment I'm leaning toward doing both 2 and 3 -- i.e. implement this as a second arg passed to `Rollbar.scope`, and make `Rollbar.scope_config(x)` a shorthand for `Rollbar.scope(nil, x)`. 

Pinging @coryvirok as well.

ptrln:
Thanks @jondeandres and @brianr.

After reading the docs a bit, it seems that I can also just call Rollbar.configure() in a before_filter or something, and change the access_token there. This should only affect the current request, and I wouldn't even need to change it back. Does that sound right?

jondeandres:
@ptrln that's viable solution yes.

I'm thinking that maybe it'll be better use `Rollbar.notifier.configure`. On each request we instantiate a new `Rollbar::Notifier` with the global configuration, but then, that notifier has its own configuration. Also, `Rollbar.notifier` return a different instance per thread. So yes, you can do something like:

```ruby
before_filter do 
  Rollbar.notifier.configure do |config|
    config.access_token = whatever_method_that_returns_the_correct_token
  end
end
```

This will not change the global configuration returned by `Rollbar.configuration`.

Let us know if this works for you :smile:.

Thanks!

antek-drzewiecki:
Yep great feature, I was actually having the same issue. 
I was generating access tokens per module that required logging.
In most cases i spawned notifiers in around_actions like this.
```
class MyModule::ApplicationController
  around_action :set_logging_scope
  
  def set_logging_scope   
      yield
    rescue => e
      notifier.critical(e)   
  end
  def notifier
    notifier = Rollbar.scope(some extra scopes if you want)
    notifier.configuration.access_token = "xxxxxxxx"
    notifier
  end
end

class MyModule::SomeController  < MyModule::ApplicationController
 def index
 end
end
```

And for gems like [grape](https://github.com/ruby-grape/grape) I would use create an own logger which overrides the access tokens.

```
module Api
  class RollbarScopedLogger < Rollbar::Notifier
    def initialize(token)
      super(Rollbar)
      self.configuration.access_token = token
    end
  end
end
```
But your way seems much cleaner.

The only issue I am heaving is with Sidekiq. @jondeandres , since you say Rollbar spawns a new notifier each threat. Then the most elegant solution would be creating Sidekiq middleware that would call `Rollbar.notifier.configure` correct?

coryvirok:
@brianr I like option 2 the most. Also, I think we should also return the Notifier instance from `configure()` so you can chain method calls.

jondeandres:
@antek-drzewiecki responding to your Sidekiq doubt:

the solution is what you've proposed, just write a middleware for Sidekiq so it configures the thread notifier with `Rollbar.notifier.configure`.

antek-drzewiecki:
I've created this [Sidekiq middleware gist](https://gist.github.com/antek-drzewiecki/142f1bbb068712b141e0) that configures the Rollbar notifier with an unique token.

jondeandres:
@antek-drzewiecki awesome, thanks!

Issue #273, opened 2015-08-13
rollbar-rails-runner works not well with rails 3.x and 4.0


ctokoro:
I've got exception in execute ```rollbar-rails-runner``` with code which contains define method.
this is sample code.
            
#### test.rb
```ruby
def say
  puts "HELLO"
end

say
```

#### rails runner
```
$ cd rails32_application

$ bundle exec rails r test.rb 
HELLO

$ bundle exec rollbar-rails-runner test.rb
test.rb:5:in `eval_runner': undefined local variable or method `say' for Rails:Module (NameError)
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:94:in `eval'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:94:in `eval_runner'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:41:in `module_eval'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:41:in `eval_runner'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:30:in `block in run'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:47:in `rollbar_managed'
	from /private/var/www/rollbar-gem/lib/rails/rollbar_runner.rb:30:in `run'
	from /private/var/www/rollbar-gem/bin/rollbar-rails-runner:5:in `<top (required)>'
	from /private/var/www/rails32_application/.bundle/bin/rollbar-rails-runner:16:in `load'
	from /private/var/www/rails32_application/.bundle/bin/rollbar-rails-runner:16:in `<main>'
```

i checked it works fine with rails 4.1 and later, and I found it is fixed by above commit of rails.

load the file rather than evaling · rails/rails@bc8eaf0
https://github.com/rails/rails/commit/bc8eaf0f5b90d0af5fa996649a5133953dd45086

but im using rails 3.2,  could you check it or give me any suggestion?

jondeandres:
Hi @ctokoro,

could you change test.rb so it has some module namespace? Ex:

```ruby
# test.rb
module MyModule
  extend self

  def say
    puts "HELLO"
  end
end

MyModule.say
```

If you see at the runner code in Rails you can see that provide this functionality is too hacky, so probably it can break in some scenarios. Really rails runner is not prepared for this.

Try this and let us know if it works.

[...]

PR #280, opened 2015-08-18
use `instance_eval` instead of `module_eval`

kibitan:
related with #273 

`rails runner` execute the file with `Rails` module in rails 3.*, 4.0 and the difference with `module_eval` and `instance_eval` is context of evaling code.
- module_eval
  - it will execute with `eigenclass` of `self` context. inside of eval code `def` will define class method.
- instance_eval
  - it will execute with `class` of `self` context. inside of eval code `def` will define instance method.

### DIFF
--- a/spec/bin/rollbar_runner_spec.rb
+++ b/spec/bin/rollbar_runner_spec.rb
@@ -13,23 +13,9 @@ describe Rails::RollbarRunner do
   context 'with file contains define method in top level' do
     # refs: https://github.com/rollbar/rollbar-gem/issues/273
 
-    before do
-      with_dummyapp_dir do
-        File.open('./lib/test_ruunner.rb', 'w') do |file|
-          file.write <<-EOS
-def hello
-  puts 'world'
-end
-
-hello
-EOS
-        end
-      end
-    end
-
     it 'do not raise exceptions' do
       with_dummyapp_dir do
-        expect( `rollbar-rails-runner ./lib/test_ruunner.rb` ).to eq "world\n"
+        expect( `bundle exec rollbar-rails-runner ./lib/test_ruunner_with_define_method_in_top_level.rb` ).to eq "world\n"
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/bin/rollbar_runner_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+end
+
+describe Rails::RollbarRunner do
+  context 'with file contains define method in top level' do
+    # refs: https://github.com/rollbar/rollbar-gem/issues/273
+
+    before do
+      with_dummyapp_dir do
+        File.open('./lib/test_ruunner.rb', 'w') do |file|
```
jondeandres: It'd be nice having this file written in the dummyapp instead of write it for every test.

jondeandres:
Hey @ctokoro, thanks for this!

Probably it has more sense use `#instance_eval` instead of `#module_eval`, yes. I'm looking there are several builds failing in travis.

* Some of them are failing cause the output is not the expected, https://travis-ci.org/rollbar/rollbar-gem/jobs/76092944.
* Others fail cause `byebug` gem is not found, https://travis-ci.org/rollbar/rollbar-gem/jobs/76092939. I think this is happening cause when you execute `rollbar-rails-runner ./lib/test_ruunner.rb` maybe the environment is not used and the `Gemfile` in the repo root is being used. Instead of the defined by `BUNDLE_GEMFILE`.
  * Another option is that you should execute that command with `bundle rollbar-rails-runner`.

ctokoro:
@jondeandres 
thanks your review and kindly advice!
i will try to fix it :smile:

brianr:
@ctokoro Do you think you'll get to this relatively soon?

ctokoro:
@brianr CC: @jondeandres 
Actually not yet. And I've found another problem with using `#instance_eval` currently.
with `include module` in runner script, module will be not included well.. (`NoMethodError` is happened.)
I've not research definitely reproduce conditions of it yet, but now I think this PR is not good solution of #273 .

coryvirok:
@rokob can you take a look and see if there's a good solution?

rokob:
Related to some other rails-runner issues #551, will take a look soon

rivkahstandig3636:

ArturMoczulski:
Issue #273 has been closed and this PR is conflicting with the current master. If this is still a requested feature I will reopen.

PR #282, opened 2015-08-27
Reraise exceptions on Rollbar.process_payload_safely.

jondeandres:
We will reraise exceptions in `Rolbar.process_payload_safely` so async queues can retry the job or, in general, handle an error report some way.

At same time that exception is silenced so we don't generate infinite reports. This example is what we want to avoid:

1. New exception in a the project is raised
2. That report enqueued to Sidekiq queue.
3. The Sidekiq job tries to send the report to our API
4. The report fails, for example cause a network failure,
   and a exception is raised
5. We report an internal error for that exception
6. We reraise the exception so Sidekiq job fails and
   Sidekiq can retry the job reporting the original exception
7. Cause the job failed and Sidekiq can be managed by rollbar we'll
   report a new exception.
8. Go to point 2.

We'll then push to Sidekiq queue indefinitely until the network failure is fixed.

Using Rollbar.silenced we avoid the above behavior but Sidekiq will have a chance to retry the original job.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -27,7 +27,7 @@ module Rollbar
     Rack::Multipart::UploadedFile
   ].freeze
   PUBLIC_NOTIFIER_METHODS = %w(debug info warn warning error critical log logger
-                               process_payload process_payload_safely scope send_failsafe log_info log_debug
+                               process_payload process_from_async_handler scope send_failsafe log_info log_debug
                                log_warning log_error silenced)
 
   class Notifier
@@ -218,7 +218,7 @@ module Rollbar
     #
     # Using Rollbar.silenced we avoid the above behavior but Sidekiq
     # will have a chance to retry the original job.
-    def process_payload_safely(payload)
+    def process_from_async_handler(payload)
       Rollbar.silenced do
         begin
           process_payload(payload)
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -209,7 +209,7 @@ module Rollbar
     # 5. We report an internal error for that exception
     # 6. We reraise the exception so Sidekiq job fails and
     #    Sidekiq can retry the job reporting the original exception
-    # 7. Cause the job failed and Sidekiq can be managed by rollbar we'll
+    # 7. Because the job failed and Sidekiq can be managed by rollbar we'll
     #    report a new exception.
     # 8. Go to point 2.
     #
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -17,7 +17,7 @@ require 'rollbar/logger_proxy'
 require 'rollbar/exception_reporter'
 require 'rollbar/util'
 require 'rollbar/railtie' if defined?(Rails::VERSION)
-require 'rollbar/delay/girl_friday'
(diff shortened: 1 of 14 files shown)


lib/rollbar.rb
```diff
@@ -195,10 +195,39 @@ def process_payload(payload)
...
+    #
+    # 1. New exception in a the project is raised
+    # 2. That report enqueued to Sidekiq queue.
+    # 3. The Sidekiq job tries to send the report to our API
+    # 4. The report fails, for example cause a network failure,
+    #    and a exception is raised
+    # 5. We report an internal error for that exception
+    # 6. We reraise the exception so Sidekiq job fails and
+    #    Sidekiq can retry the job reporting the original exception
+    # 7. Cause the job failed and Sidekiq can be managed by rollbar we'll
```
brianr: Typo: s/Cause/Because/

lib/rollbar/delay/resque.rb
```diff
@@ -23,7 +23,12 @@ def self.perform(payload)
         end
 
         def perform(payload)
-          Rollbar.process_payload_safely(payload)
+          begin
+            Rollbar.process_payload_safely(payload)
+          rescue
+            # Raise the exception so Resque can track the errored job
```
brianr: Is this begin-rescue necessary? Seems like the code would function the same if it was just:

```ruby
def perform(payload)
  Rollbar.process_payload_safely(payload)
end
```

lib/rollbar.rb
```diff
@@ -195,10 +195,39 @@ def process_payload(payload)
...
+    #    report a new exception.
+    # 8. Go to point 2.
+    #
+    # We'll then push to Sidekiq queue indefinitely until the network failure
+    # is fixed.
+    #
+    # Using Rollbar.silenced we avoid the above behavior but Sidekiq
+    # will have a chance to retry the original job.
     def process_payload_safely(payload)
```
brianr: This method name feels potentially misleading -- i.e. what exactly does `safely` mean? I would suggest we call it something more descriptive of what it does, maybe `process_payload_inside_async_handler`

lib/rollbar.rb
jondeandres: :+1:

lib/rollbar/delay/resque.rb
jondeandres: I wanted to be explicit here for the reraise. I can remove it and change the comment. Better for you?

lib/rollbar.rb
jondeandres: Yah...the name was safely cause it swallowed the exceptions before this change. I'll change the name.

lib/rollbar/delay/resque.rb
brianr: Either way is fine with me. The comment is nice :)

lib/rollbar.rb
jondeandres: I've renamed the method in https://github.com/rollbar/rollbar-gem/commit/2ea86655a8d8c79d7d9857b8015fe0a05e053a15.

dgilperez:
:clap:

PR #283, opened 2015-08-28
Send class name on internal failsafe messages.

jondeandres:
This will help us with the future errors and customer support.


brianr:
:+1:

PR #286, opened 2015-09-01
Only resolve encoding_class once

dvydra:
Repeated invocations of either require or instance_methods seems to
consume a lot of memory.

We’re ok to just support the encoder as a test.

### DIFF
--- a/lib/rollbar/encoding.rb
+++ b/lib/rollbar/encoding.rb
@@ -1,3 +1,5 @@
+require 'rollbar/encoding/encoder'
+
 module Rollbar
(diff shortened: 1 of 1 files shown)


brianr:
Thanks @dvydra !

@jondeandres has done some work on this as well, over here: https://github.com/rollbar/rollbar-gem/pull/286 . I expect we'll get one/both of these merged within the next day.

dvydra:
Yep, I will be testing #286 today, this was intended to be a PR on our fork only, this one can be closed, it's not suitable for the general public.

PR #289, opened 2015-09-01
Send session data instead of session store options.

jondeandres:
We are sending session options instead of session data. I really don't
know the reason cause anybody could be interested on this instead of
request session data.

Session options in a Rails application can be something like this:

```ruby
Your::Application.config.session_store :cookie_store, {
  :key =>           '_session_id',
  :path =>          '/',
  :domain =>        nil,
  :expire_after =>  nil,
  :secure =>        false,
  :httponly =>      true,
  :cookie_only =>   true
}
```

Which is, in general, static, doesn't depend on the current request and
well, customers can always just see their configuration file.

I'd say the original intention was send session data.

This fix the serializing JSON problem we are having in some cases where there are socket objects
to Redis or other store backends in those session options.


jondeandres:
This, probably, doesn't fix completely the problem we are having serializing JSON in some scenarios but avoid it when is caused by session options structure.

brianr:
lgtm

PR #290, opened 2015-09-04
Send nearest backtrace entry on failsafe messages. @brianr.

jondeandres:


lib/rollbar.rb
```diff
@@ -619,6 +605,33 @@ def send_failsafe(message, exception)
       failsafe_payload
     end
 
+    def failsafe_body(message, exception)
+      body = 'Failsafe from rollbar-gem. '
+      log_error "[Rollbar] Sending failsafe response due to #{message}."
```
brianr: This log message should probably still be in `send_failsafe`

lib/rollbar.rb
```diff
@@ -619,6 +605,33 @@ def send_failsafe(message, exception)
...
+
+          log_error "[Rollbar] #{body}"
+        rescue
+        end
+      else
+        begin
+          body += message.to_s
+        rescue
+        end
+      end
```
brianr: Couple things to avoid a potential 413 or 422 when the request is made to our API:

- Should truncate the resulting body in case it's too long
- Any chance of encoding issues?

jondeandres: @brianr this payload is passed through the dump method that will truncate it according to our strategies.

About the encoding. the message actually have:

- Our own part of message: 'Failsafe from rollbar-gem. '
- The exception class name. This doesn't have exception message value, it'll be safe.
- The nearest entry from backtrace. Here we could send rare chars to the API if the customer has some directory names with those chars, what will not be usual. But in this case I think is safer don't overuse the encoding logics we have so we don't add more potential risk to the safely report.

lib/rollbar.rb
jondeandres: https://github.com/rollbar/rollbar-gem/commit/e81d5fccfa23b6d8b318c8409bf1f610f300ba35

Issue #291, opened 2015-09-07
Incorrect behaviour in DelayedJob


ka8725:
I recently noticed that Rollbar doesn't work in my DelayedJob tasks. Tried to debug: just placed `raise 'h'` in the delayed method and run `Delayed::Job.last.invoke_job` from rails console. This is the result:

```
[Rollbar] Scheduling payload
[Rollbar] Sending payload
[Rollbar] Got unexpected status code from Rollbar api: 422
[Rollbar] Response: {
  "err": 1,
  "message": "Invalid format. data.request string value found, but a object is required. data.request an object is required."
}
[Rollbar] Details: https://rollbar.com/instance/uuid?uuid=a351fb56-2ee9-4628-88d6-d7a735051771 (only available if report was successful)
RuntimeError: h
```

Any ideas what is wrong?

ka8725:
More info about this bug: it works with `report_dj_data = false` option set.

ri4a:
Pull request #292 fixes this.

brianr:
Thanks @ka8725 and @ri4a!

What version of delayed_job were you seeing this in? I am fairly confident that this has worked in the past, so I'm guessing it may be broken in the latest version / some other version than we are already testing in. We'll want to update the test suite to make sure it covers this.

ka8725:
I use `delayed_job_active_record-4.0.3` Ruby 2.1.5

ri4a:
It doesn't look like a delayed_job issue. Last time we received a delayed_job error was right before we upgraded from 1.5.2. to 2.1.1...

Looks like the problem was introduced when the enforce_valid_utf8 method in /lib/rollbar.rb was changed from

```ruby
def enforce_valid_utf8(payload)
  normalizer = lambda do |object|
    is_symbol = object.is_a?(Symbol)
    return object unless object == object.to_s || is_symbol
    # ...removed some code here...
  end
  Rollbar::Util.iterate_and_update(payload, normalizer)
end
```

to

```ruby
def enforce_valid_utf8(payload)
  normalizer = lambda { |object| Encoding.encode(object) }
  Rollbar::Util.iterate_and_update(payload, normalizer)
end
```

Note the

```ruby
return object unless object == object.to_s
```

in the old method, whereas in the new one the object is immediately converted to String:

```ruby
def encode
  value = object.to_s
  encoding = value.encoding

  # This will be most of cases so avoid force anything for them
[...]

Issue #94, opened 2014-02-16
Failsafe from rollbar-gem: Could not send payload due to it being too large after truncating attempts


lukes:
Hi, 

We're seeing these errors appear now after upgrading from 11.7 to 12.10. We see around 6-7 occurrence per day in production. 

I only raise this in case it's helpful for you to know that at least someone if encountering this error relatively frequently, and perhaps it's possible to continue to attempt to reduce the payload, or perhaps the method of reduction could be looked at.

Looking at our log, it appears as though the payload size is due to the stacktrace. It may be beneficial to attempt to prune the stacktrace.

jasper-fu:
![image](https://cloud.githubusercontent.com/assets/233379/8642176/d2e8481c-2952-11e5-9ccb-8daa0437666e.png)
Hey, guys, I keep seeing this, what should I do to fix this? thanks

jondeandres:
Hi @jasper-fu, we'll release a new version today or tomorrow. There are some scenarios where active support is breaking the serialization to JSON.

We have a PR for this, https://github.com/rollbar/rollbar-gem/pull/254. But needs to be tested yet.

Sorry for the inconveniences.

jondeandres:
@jasper-fu are you using redis-session-store in your project?

if so, you can ensure that you are configuring the Rails session-store without passing a redis instance, and passing a hash options with the redis initialization options.

jasper-fu:
@jondeandres ok, thanks for helping. :)

exterm:
We have this problem again. Maybe the rollbar guys can check it at https://rollbar.com/Avocadostore/Avocadostore/items/430/ . Is there some way to get more details on the error?

jondeandres:
@exterm are you using Rails? Can you find messages like `[Rollbar]` on logs?

Classname and a backtrace should appear in the logs.

Could you write us `support@rollbar.com`?

LucasPadovan:
We are having the same issue again, this error appears almost every week for any version of the gem, can we do something to deactivate this message? Rollbar is charging us because we have 2k+ of these occurrences...

brianr:
@LucasPadovan Thanks for the report... can you find the `[Rollbar]` messages in your logs that have the details about these errors, and send them to us at `support@rollbar.com`?

[...]

PR #292, opened 2015-09-08
make delayed_job exceptions work (issue 291)

ri4a:


jondeandres:
@ri4a thanks for this, I guess this fix https://github.com/rollbar/rollbar-gem/issues/94#issuecomment-138386657 ?

I'll test this and merge if everything is ok. I think we'd do better using `Rollbar::JSON.dump` interface, but the fix looks reasonable.

Thanks!

jondeandres:
@ri4a sorry, this PR doesn't fix https://github.com/rollbar/rollbar-gem/issues/94.

I'm looking at this, thanks again for the PR.

jondeandres:
@ri4a 

I've looked at your comments in https://github.com/rollbar/rollbar-gem/issues/291#issuecomment-138730134.

We removed this part of code:

```ruby
is_symbol = object.is_a?(Symbol)

return object unless object == object.to_s || is_symbol
```

But we are doing this in `lib/rollbar/encoding.rb`:

```ruby
def self.encode(object)
    can_be_encoded = object.is_a?(String) || object.is_a?(Symbol)

    return object unless can_be_encoded

    encoding_class.new(object).encode
end
```

I'd say it's basically the same, maybe I'm missing something. The goal there is only encode `String`  and `Symbol` objects.

Having `return object unless object == object.to_s` is just a way to check it is a `String` object.

Are you sure the problem is here? It'd be nice having a test that exploits the current bug.

Thanks.

ri4a:
@jondeandres

Yes, you are right. I looked at Rollbar::Encoding::Encoder#encode instead of Rollbar::Encoding::encode :unamused:

The real cause is in /lib/rollbar/truncation/mixin.rb, which changed from

```ruby
def dump(payload)
  MultiJson.dump(payload)
end
```

to

```ruby
def dump(payload)
  Rollbar::JSON.dump(payload)
end
```

Apparently the serialization of Ruby objects is different (seems that now only to_s is called).

To create a test that exploits the bug, it depends on what you consider to be the actual bug here. The fact that a method called "enforce_valid_utf8" is called with unserialized Ruby objects seems strange, since the serialization afterwards could cause new encoding issues. Should Rollbar::Encoding::encode really accept (i.e. ignore) Ruby objects and trust they will be serialized correctly later by Truncation#dump? Or should it be the responsibility of the data injectors to first serialize? Or maybe "enforce_valid_utf8" should just be called later.

This PR fixes #291 by serializing the Job object where it is injected. It could also be fixed by amending Rollbar::JSON. That may or may not also fix #94, haven't looked at that.

jondeandres:
@ri4a that has more sense. the `enforce_valid_ut8` method is not the problem here. I agree with you that maybe we should enforce utf8 after dump the payload. The thing here is that we had some problems with some srings when an async handler pushed to its queue, so we enforce utf8 before.

Apart of this, and focused in the bug, we should probably pass `request: job.payload_object.as_json` from the beginning and not a delayed job `Job` instance.

I've opened a PR for this, https://github.com/rollbar/rollbar-gem/pull/293.

I think we can close this one, but anyway, thanks for open it!

PR #293, opened 2015-09-10
Fix delayed_job crash reports. Fixes #291

### LINKED ISSUE #291: Incorrect behaviour in DelayedJob

jondeandres:
Since serialization is made by JSON, we crashed delayed job reports
cause we were exploting the JSON & ActiveSupport bug. Now we get the job
data from job.payload_object.as_json.

I've added tests for this, testing two different scenarios:

- method without arguments
- method with arguments


brianr:
lgtm

PR #295, opened 2015-09-11
Remove whitespace from config template

levent:
Lots of apps using this template will have style guide analysers (like rubocop) that complain about this sort of minor thing.


jondeandres:
awesome @levent, thanks!

PR #298, opened 2015-09-13
JSON.generate returns NoMethodError key?, need to handle exception an…

### LINKED ISSUE #296: Failsafe from rollbar-gem:  [ERROR] error in process_payload without stack trace
Failsafe from rollbar-gem: NoMethodError: error in process_payload is our specific error. Upon further investigation inside of rollbar, I found this was due to using JSON.generate(object, options) instead of using object.to_json

JSON.generate only allows objects or arrays to be converted to JSON syntax. to_json, however, accepts many Ruby classes even though it acts only as a method for serialization.

Using generate returns a NoMethodError key? in some instances. 

I will submit a PR on this.

heshamnaim:
…d try to_json for activerecord

Closes #296 

JSON.generate only allows objects or arrays to be converted to JSON syntax. to_json, however, accepts many Ruby classes even though it acts only as a method for serialization. Need to capture exception from JSON.generate and try to_json in that instance to handle some activerecord objects

### DIFF
--- a/lib/rollbar/json.rb
+++ b/lib/rollbar/json.rb
@@ -26,7 +26,18 @@ module Rollbar
     end
 
     def dump(object)
-      dump_method.call(object)
+      # JSON.generate defined above returnes a NoMethodError for key? for some activerecord definitions
+      begin
+        dump_method.call(object)
+      rescue => e
+        # If exception caught, try to_json 
+        result = object.try(:to_json)
+        if result.nil?
+          raise e
+        else
+          result
+        end
+      end
     end
 
     def load(string)


jondeandres:
@heshamnaim I've replied you in https://github.com/rollbar/rollbar-gem/issues/296.

We decided avoiding ActiveSupport use cause there are some bugs there serializing JSON, like infinite loops or bad performance for some payloads. And was easier having all the people using the same library.

This problem comes because conflicts between JSON and ActiveSupport, we are thinking which solution adopt for this.

heshamnaim:
@jondeandres yeah i figured, thats why I only capture if JSON.generate doesn't work and try. We monkey patched it this way in our project and now get the full proper error.

jondeandres:
@heshamnaim this is suposed to be fixed inhttps://github.com/rollbar/rollbar-gem/blob/master/CHANGELOG.md#230.

We've changed few things to improve the JSON serialization. If you have still problems please write us at support@rollbar.com.

Thanks!

PR #300, opened 2015-09-15
Load Oj instead of JSON.

jondeandres:


lib/rollbar.rb
```diff
@@ -796,14 +796,7 @@ def require_core_extensions
     end
 
     def monkey_patch_socket?
-      return false unless defined?(ActiveSupport::VERSION::STRING)
```
jondeandres: Testing failed with active support 4.2. So just monkey patch always active support is loaded.

PetrKaleta:
You can't use Oj omg as its C extension gem! So what about JRuby omg

lib/rollbar.rb
PetrKaleta: I am just curious how this change is related with this issue

jondeandres: @PetrKaleta ActiveSupport has some bugs when calling `#to_json` or `#as_json`. Most of the JSON serialization libraries calls somewhere `#as_json` if it's defined.

We've experienced many customers that in their payloads, for some reason, have objects that internally have an instance of `BasicSocket`. ActiveSupport then arrives to an infinite loop trying to execute `#as_json`. This caused by the monkey patch they add to `Object`.

And in this PR we realized we were having problems with all the ActiveSupport versions.

PetrKaleta: Thanks for explanation, makes sense

PR #304, opened 2015-09-18
Restore MultiJson and add custom option modules. Fix #303

### LINKED ISSUE #303: v2.3.0 and what bout JRuby?
In 2.3.0 you're using Oj gem. Ok, fine, and what about JRuby? You can't use globally gem with C extensions...

jondeandres:
In order to support JRuby it'd be better not force installing Oj since
it cannot be used in JRuby if C extensions are not enabled, and they are
not supported by default.

So, we add MultiJson again to the project. This will create more
different scenarios but since it could be the best option according to
previous paragraph.

We've added in this change support for custom options depending on the
loaded adapter in MultiJson. We can add more adapter options modules in
this way:

```ruby
module Rollbar
  module JSON
    module AnyMultiJsonAdapterName
      def self.options
        { # here comes the options we need for that adapter }
      end
    end
  end
end
```

We have, for now, just the Oj module that will return the options we had
before this commit.


joneslee85:
Not a big fan of this PR but pragmatically it is :+1: to do

PR #306, opened 2015-09-21
Change README.md and template to use 'default' Sidekiq queue

jondeandres:


albertyw:
I think the text of this PR's title and the changelog (https://github.com/rollbar/rollbar-gem/blame/master/CHANGELOG.md#L17) should be clarified, perhaps to `Change README.md and template to show example of using 'default' Sidekiq queue`.  My first thought reading the changelog was that rollbar would switch the default sidekiq queue to 'default'

jondeandres:
@albertyw probably you are right yes, we can change the CHANGELOG.md text to be more clear.

PR #310, opened 2015-09-28
Don't require rollbar/rails on initializer

jondeandres:
Rollbar::Rails becomes an empty module. That functionality is given by
the Railtie.

However, seems that in the tests we were calling
`Rollbar::Rails.initialize`, so the method has been moved to a spec helper.

### DIFF
--- a/lib/generators/rollbar/templates/initializer.rb
+++ b/lib/generators/rollbar/templates/initializer.rb
@@ -1,4 +1,3 @@
-require 'rollbar/rails'
 Rollbar.configure do |config|
   # Without configuration, Rollbar is enabled in all environments.
   # To disable in specific environments, set config.enabled=false.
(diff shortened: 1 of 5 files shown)


lib/generators/rollbar/templates/initializer.rb
```diff
@@ -1,4 +1,3 @@
-require 'rollbar/rails'
```
brianr: Will customers that already have this line in their generated initializer need to remove it? If they don't remove it, what will happen?

jondeandres: The file has not been deleted, so the require itself doesn't generate any error. That file just configured the notifier for rails, `framework`, `logger`, etc...But that is actually done in the railtie, what is enough.

PR #312, opened 2015-09-30
Allow custom revision name on deploy

fabn:
If one wants to use tags or anything else to name revisions it would be nice to have the option to pass it for rollbar deployments.


jondeandres:
@fab the PR looks good. we'll merge it soon.

Thanks!

fabn:
:+1:  you're welcome.

PR #313, opened 2015-10-01
Remove better errors hook

jondeandres:
We have our own middleware reporting errors to the API. Monkey patching
better_errors in the practice the behavior is just reporting the errors
twice: one by our middleware and another one by better_errors.

Beside this, depending on the better_errors version we have two
different behaviors:

- versions < 2.0: the error is reported twice
- versions >= 2.0: we report one error by our middleware and a 'Empty  message' error cause the exception managed by better_errors doesn't inherit `Exception` and it's a custom object by better_errors.

I'm not sure why we had this monkey patch but seems it's not useful and
in any scenario is becoming finally in two reports by exception.


coryvirok:
Makes sense. Will this cause any users to need to modify their code to continue getting Rollbar errors reported? And is this change backward compatible with people's current install of rollbar-gem?

jondeandres:
@coryvirok no, users that have better_errors installed just will see their errors, in development, are sent just once. now we are sending them twice.

PR #316, opened 2015-10-07
Verify peer when connecting to Rollbar over HTTPS

mcmillan:
At the moment, when connecting over HTTPS to Rollbar (which is the default, rightly so) and `use_eventmachine` is set to `false` (the default), `Net::HTTP`'s `verify_mode` is set to `OpenSSL::SSL::VERIFY_NONE`. This effectively means that the Rollbar gem doesn't validate whether or not the Rollbar endpoint it's talking to has a valid SSL certificate. This makes the Rollbar gem susceptible to MITM attacks.

The implications of this are obviously significant; sensitive data is often (intentionally or otherwise) transmitted as part of exception data which could be intercepted by a malicious third-party and tampered with. If a developer has Rollbar notifications switched on in their development environment and they're building a sweet new application in a coffee shop with insecure wifi, it's perfectly possible for an attacker to steal their exception data and do something nasty.

This changes the `verify_mode` to be `OpenSSL::SSL::VERIFY_PEER`, which will verify that the Rollbar endpoint that data is being sent to has a valid SSL certificate before transmitting data across.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -528,7 +528,7 @@ module Rollbar
 
       if uri.scheme == 'https'
         http.use_ssl = true
-        http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+        http.verify_mode = OpenSSL::SSL::VERIFY_PEER
       end
 
       request = Net::HTTP::Post.new(uri.request_uri)


brianr:
Thanks @mcmillan . Agreed, we should do this.

Looks like this changes causes the tests to fail on many of the platforms: https://travis-ci.org/rollbar/rollbar-gem/builds/84016192 . Any chance you know how to fix some of those?

jondeandres:
@brianr there are a couple of failures that are caused by bundle installation

but https://travis-ci.org/rollbar/rollbar-gem/jobs/84016195 seems to be failing by other reason

mcmillan:
@brianr As @jondeandres mentions I suspect the errored builds are caused by a version of `rack-cache` somewhere that isn't properly pinned. `rack-cache` 1.3.0 [was released](https://github.com/rtomayko/rack-cache/releases/tag/v1.3.0) a few days ago and requires Ruby > 2.0.0, hence the erroring builds.

As for the failing build – I have no idea what's going on there, to be honest. Can you guys advise?

jondeandres:
hey @mcmillan, I can take a look at this. I hope soon.

coryvirok:
Thanks for reporting, @mcmillan. We will get a fix for this out this coming week.

PR #317, opened 2015-10-08
Fix when MultiJson is using adapters other than Oj

godfat:
Whenever Yajl is used for MultiJson, before this patch,
Rollbar would find ::Yajl instead of Rollbar::JSON::Yajl,
and ::Yajl does exist but it does not respond to `options`,
causing failures.

It's surprising that Rollbar::JSON.const_get(:Yajl) would
find ::Yajl instead, it's like using Rollbar::JSON::Yajl to
refer to ::Yajl, this is really unfortunate for Ruby.
But given current status, we cannot rely on const_get in
this case. Therefore we'll need to search on constants
instead.

This has another advantage -- we no longer need to rescue,
saving an exception.

### DIFF
--- a/lib/rollbar/json.rb
+++ b/lib/rollbar/json.rb
@@ -43,7 +43,7 @@ module Rollbar
     def find_options_module
       module_name = multi_json_adapter_module_name
 
-      if constants.find{ |const| const.to_s == module_name }
+      if const_defined?("::Rollbar::JSON::#{module_name}")
         const_get(module_name)
       else
         Default
--- a/lib/rollbar/json.rb
+++ b/lib/rollbar/json.rb
@@ -43,9 +43,9 @@ module Rollbar
     def find_options_module
       module_name = multi_json_adapter_module_name
 
-      begin
+      if constants.find{ |const| const.to_s == module_name }
         const_get(module_name)
-      rescue NameError
+      else
         Default
       end
     end
--- a/spec/rollbar/json_spec.rb
+++ b/spec/rollbar/json_spec.rb
@@ -24,6 +24,11 @@ module MultiJson
   end
 end
 
+module MissingCustomOptions
+  # Consider the fact that there's MultiJson::Adapters::Yajl but not
+  # Rollbar::JSON::Yajl, it should not look for ::Yajl but only
+  # Rollbar::JSON::Yajl.
+end
 
 describe Rollbar::JSON do
   let(:payload) do


jondeandres:
@godfat good catch 

I think I'd feel more comfortable using `const_get("::Rollbar::JSON::#{module_name}")` instead find on constants. What do you think?

godfat:
As long as it works, I am ok with it, but does it work?

```
>> Rollbar::JSON.const_get("Rollbar::JSON::Yajl")
=> Yajl
```

jondeandres:
@godfat `Object.const_get("::Rollbar::JSON::Yajl")`

If this doesn't work then use `constants` may be ok, but I don't like that `find` too much.

godfat:
Alternatives:

``` ruby
constants.include?(module_name.to_sym)
# I would try to avoid #to_sym though, and old rubies
# could return a list of strings rather than symbols
```

``` ruby
constants.map(&:to_s).include?(module_name)
# An extra array and complexity: 2n
```

godfat:
Oh, change `find` to `any` should also work:

``` ruby
constants.any?{ |const| const.to_s == module_name }
```

jondeandres:
@godfat I've been trying with this snippet:

```ruby
module A
end

module B
  extend self

  def get_a
    puts const_get('::B::A')
  end

  def is_defined_a
    puts const_defined?('::B::A')
  end
end


B.get_a
B.is_defined_a

Output:
A
false

```

I'd like use `const_defined?` instead look at `constants`, any opinion about this?

Thanks!

godfat:
It seems `const_defined?` with full name should work:

```
>> Rollbar::JSON.const_defined?("Yajl")
=> true
>> Rollbar::JSON.const_defined?("::Rollbar::JSON::Yajl")
=> false
>> Rollbar::JSON.const_defined?("::Rollbar::JSON::Default")
=> true
```

So I think that could be the best choice given current status.
Thanks!
Do you want me to commit and optionally rebase, or do you want to apply the change?

jondeandres:
@godfat if you are interested  on fix the PR we'll be happy to merge it. If you don't find a moment to do it we'll do soon :smile: 

Thanks!

godfat:
Great! I've committed. I am eager to fix this right now, thank you for the immediate help!

jondeandres:
Awesome! When the tests pass we can merge it.

Thanks!

godfat:
My change in `spec/rollbar/json_spec.rb` would reveal this, so I think it's good enough. It won't hurt to have another test case though, of course.

jondeandres:
hey @godfat, may tests are failing in travis. if you want to take a look at them it'd be nice to have this merged soon. if you don't find a time to it we'll analyze them ourself.

Thanks.

godfat:
I don't see why they were failing and it's a bit random to me. It looks like you're running a lot of tests with the same ruby version and one could pass but not the other one. So I would guess it's a Travis thing and maybe it would fix itself by rerunning them. (I am not very sure if const_defined? would work for older rubies though)

I would also suggest to reduce the combination of options on Travis because a lot of rubies look exactly the same.

Thanks for analyzing the tests.

fabn:
I tried RB 2.4.0 in a project with Yajl and sidekiq rollbar flooded my sidekiq rollbar queue with thousands of errors with message

```
undefined method `options' for Yajl:Module
```

I stopped my sidekiq worker and manually removed all those errors from queue. I think this issue is the cause, am I right?

jondeandres:
Hi @fabn, sorry about this.

This PR will fix your problem yes. We still need to take in deep a look into the code and the tests since they are failing.

We hope we can merge this and release it soon.

fabn:
Ok, I'll follow this issue. I'd suggest to put a big warning in your readme for 2.4.0 because I stopped sidekiq after 2 minutes at ~90000 errors queued. I think because this code was triggered by internal reports and every processed job enqueue a new one flooding the queue.

This can cause a DOS in your app in minutes if not immediately detected and manually stopped.

Issue #318, opened 2015-10-14
Sidekiq report threshold


pacoguzman:
Rollbar extensions includes some configuration parameters to report error since a specific job attempt when using delayed job, but not when using Sidekiq.

Would you accept a PR that solve that?

jondeandres:
@pacoguzman sure, feel free to open a new PR and we can then discuss over the code.

Thanks!

pacoguzman:
@jondeandres I just created one #319

jondeandres:
awesome :smile:

aselder:
There is a big issue with this PR: https://github.com/rollbar/rollbar-gem/issues/345

jondeandres:
This can be closed.

PR #319, opened 2015-10-14
check threshold before a job is reported to Rollbar

pacoguzman:
This PR implements a simple `Sidekiq` threshold similar to the existing for `DelayedJob` (#318).

If you want to reduce the noise of your `Rollbar` inbox for `Sidekiq` jobs that fails only on the first attempt but not the next ones you can define in the Rollbar configuration this threshold so only notify from a number of retry of the current failing job.

By default is disabled the threshold is 0. And for jobs that are not going to be retried we notify always to Rollbar.

Any feedback is welcome, thanks in advance

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -5,10 +5,16 @@ module Rollbar
     PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
 
     def self.handle_exception(msg_or_context, e)
+      return if skip_report?(msg_or_context, e)
+
       params = msg_or_context.reject{ |k| PARAM_BLACKLIST.include?(k) }
       scope = { :request => { :params => params } }
 
       Rollbar.scope(scope).error(e, :use_exception_level_filters => true)
+    end
+
+    def self.skip_report?(msg_or_context, e)
+      msg_or_context.is_a?(Hash) && msg_or_context["retry"] && msg_or_context["retry_count"] < ::Rollbar.configuration.sidekiq_threshold
     end
 
     def call(worker, msg, queue)
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -35,6 +35,7 @@ module Rollbar
     attr_accessor :scrub_fields
     attr_accessor :uncaught_exception_level
     attr_accessor :scrub_headers
+    attr_accessor :sidekiq_threshold
     attr_accessor :use_async
     attr_accessor :use_eventmachine
     attr_accessor :web_base
@@ -80,6 +81,7 @@ module Rollbar
                        :confirm_password, :password_confirmation, :secret_token]
       @uncaught_exception_level = 'error'
       @scrub_headers = ['Authorization']
+      @sidekiq_threshold = 0
       @safely = false
       @use_async = false
       @use_eventmachine = false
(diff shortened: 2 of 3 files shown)


lib/rollbar/sidekiq.rb
```diff
@@ -5,6 +5,8 @@ class Sidekiq
     PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
 
     def self.handle_exception(msg_or_context, e)
+      return if msg_or_context.is_a?(Hash) && msg_or_context["retry"] && msg_or_context["retry_count"] < ::Rollbar.configuration.sidekiq_threshold
```
jondeandres: Can you move this to a new method? Something like `skip_report?` or similar.

jondeandres:
Thanks for this. Everything looks nice, just one comment in https://github.com/rollbar/rollbar-gem/pull/319#discussion_r42035970.

pacoguzman:
Opps we need to keep raising the exception in any case the job is retried, on it

pacoguzman:
Forgot my previous comment is not true this is only how we handle the exception to report to rollbar, everything seems fine. I just extracted the code to a method as you suggested

jondeandres:
@pacoguzman yes, on the `#call` method we reraise the exception, what should be still done. I think the PR is fine :smile:. We'll merge it in next days.

Thanks!

pacoguzman:
Cool you're welcome

jondeandres:
@pacoguzman one doubt about this, since now I'm not sure this is the behavior you expected.

First fail on a job will be always reported to Rollbar, since the `msg_or_context` doesn't have `retry` set to `true`.

After the first report, Rollbar reports will be disabled for repeated job until reach the threshold, 0 by default. So for a job failing 5 times and a threshold of 3 we will have:

1 fail - reported
2 fail - skipped (retry_count 1)
3 fail - skipped (retry_count 2)
4 fail - reported (retry_count 3)
5 fail - reported (retry_count 4)

Is this the expected behavior?

Thanks.

pacoguzman:
I'll check to be 100% sure, but I think Sidekiq alwyas provide retry key and when retry key is present the retry_count even for the first attempt. We saw first errors on Rollbar with that value (retry_count = 0)

I'll confirm asap

jondeandres:
Yes, Sidekiq will provide key `retry` for first attempt with value `false`. So with this logic:

` msg_or_context.is_a?(Hash) && msg_or_context["retry"] && msg_or_context["retry_count"] < ::Rollbar.configuration.sidekiq_threshold`

For the first attempt it's always false, that's ok.

My question was about if you expect to *not* skip the irst report and then skip reports until reach the threshold.

pacoguzman:
No that's not the expected behaviour sorry my fault. I'll amend the changes

pacoguzman:
I think the behaviour you described above is not what is happening. For example

```ruby
class RetryableWorker
  include Sidekiq::Worker

  sidekiq_options retry: true

  def perform
  end
end
```

1 fail - skipped (retry true)
2 fail - skipped (retry true retry_count 1)
3 fail - skipped (retry true retry_count 2)
4 fail - reported (retry true retry_count 3)
5 fail - reported (retry true retry_count 4)

```ruby
class NoneRetryableWorker
  include Sidekiq::Worker

  sidekiq_options retry: false

  def perform
  end
end
```

1 fail - reported (retry false)

And this is the expected behaviour, what do you think?

jondeandres:
mmm @pacoguzman, I think you are right yes :smile:, sorry for the confusion.

let me review this again and we'll merge it.

Thanks!

aselder:
This PR has a bug and should be fixed or reverted: https://github.com/rollbar/rollbar-gem/issues/345

PR #320, opened 2015-10-15
Fix SSL verify mode

jondeandres:


brianr:
Nice, tests are passing! Looks good to me.

Issue #314, opened 2015-10-02
Body params are skipped for Accept header


jingweno:
The code [here](https://github.com/rollbar/rollbar-gem/blob/4ef7656c463b2a13d38d4c2fd95b456e74d199c3/lib/rollbar/request_data_extractor.rb#L124) only parses the body if content type is "application/json", while typically the "Accept" header is used instead. For example, Accept=application/vnd.heroku+json; version=3. The current implementation causes a wide range of JSON request body not to be reported.

jondeandres:
@jingweno good point, we'll change this later.

thanks for the tip!

jondeandres:
Closed by https://github.com/rollbar/rollbar-gem/issues/322

PR #322, opened 2015-10-16
#314 Treat HTTP requests containing a JSON accept header as a JSON request

elight:
For the purposes of deciding if and how to treat the body of a request, respect any JSON type specified in the Accept header.

### DIFF
--- a/lib/rollbar/request_data_extractor.rb
+++ b/lib/rollbar/request_data_extractor.rb
@@ -29,11 +29,12 @@ module Rollbar
       route_params = rollbar_filtered_params(sensitive_params, rollbar_route_params(env))
 
       params =
-        if !raw_body_params.empty?
+        if json_request_missing_content_type?(env)
           raw_body_params
         else
-          request_params.merge(get_params).merge(post_params)
+          request_params.merge(get_params).merge(post_params).merge(raw_body_params)
         end
+
 
       data = {
         :params => params,
@@ -126,7 +127,7 @@ module Rollbar
       correct_method = rack_req.post? || rack_req.put? || rack_req.patch?
 
       return {} unless correct_method
-      return {} unless request_has_json_body?(rack_req)
+      return {} unless json_request?(rack_req)
 
       Rollbar::JSON.load(rack_req.body.read)
     rescue
@@ -135,9 +136,19 @@ module Rollbar
       rack_req.body.rewind
     end
 
-    def request_has_json_body?(rack_req)
+    def json_request?(rack_req)
       rack_req.env['CONTENT_TYPE'] =~ %r{application/json} ||
       rack_req.env['ACCEPT']       =~ /\bjson\b/
+    end
+
+    def json_request_missing_content_type?(env)
+      sensitive_params = sensitive_params_list(env)
+      rack_req = ::Rack::Request.new(env)
+      raw_body_params = rollbar_filtered_params(sensitive_params, mergeable_raw_body_params(rack_req))
+
+      env['CONTENT_TYPE'] !~ /application\/json/ &&
+      json_request?(rack_req) &&
+      !raw_body_params.empty?
     end
 
     def rollbar_request_params(env)
@@ -214,6 +225,7 @@ module Rollbar
       end
     end
 
+    # TODO: Maybe this can be memoized?
     def sensitive_params_list(env)
       Array(Rollbar.configuration.scrub_fields) | Array(env['action_dispatch.parameter_filter'])
     end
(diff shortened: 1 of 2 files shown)


elight:
Addresses #314. Would love to see this merged and released soon. Thanks!

lib/rollbar/request_data_extractor.rb
```diff
@@ -130,6 +135,11 @@ def rollbar_raw_body_params(rack_req)
       rack_req.body.rewind
     end
 
+    def request_has_json_body?(rack_req)
+      rack_req.env['CONTENT_TYPE'] =~ %r{application/json} ||
+      rack_req.env['ACCEPT']       =~ /\bjson\b/
```
jingweno: This regexp may not work for a typical `ACCEPT` header: `Accept: application/vnd.heroku+json; version=3`

jingweno: Never mind. `\b` matches word boundary.

dmcinnes:
Nice :+1:

lib/rollbar/request_data_extractor.rb
```diff
@@ -28,7 +28,12 @@ def extract_request_data_from_rack(env)
       session = rollbar_filtered_params(sensitive_params, rollbar_request_session(rack_req))
       route_params = rollbar_filtered_params(sensitive_params, rollbar_route_params(env))
 
-      params = request_params.merge(get_params).merge(post_params).merge(raw_body_params)
+      params =
+        if !raw_body_params.empty?
```
jondeandres: @elight this change is not related to the PR title. We may probably discuss this in another one.

jondeandres:
Thanks @elight for the PR :smile:. I left a comment https://github.com/rollbar/rollbar-gem/pull/322#discussion_r42397639, apart of that this is ok for me.

lib/rollbar/request_data_extractor.rb
elight: After making my changes, **without** L31-36 above, I eventually found that `params` was getting the post params twice: once as the desired hash and once in some serialized form attached to a key called `key`, IIRC. This seems to be because the `post_params` contained this undesirable value while the `raw_body_params` had what I needed to get the new tests passing while not breaking the old tests.

lib/rollbar/request_data_extractor.rb
elight: `\b` is a Good Thing™ ;-)

lib/rollbar/request_data_extractor.rb
jondeandres: @elight I'd say we are getting the `post_params` and `raw_body_params` cause Rack doesn't parse JSON payload when you ask for `request.POST`. But you are probably right in the sense that we are doing something wrong. We'll look in detail to this.

But anyway, I think this logic shouldn't be here in this PR. What do you think?

elight: @jondeandres The PR doesn't pass specs without this logic.

jondeandres: Hi @elight we cannot change that logic in the way you changed it. You are not adding the GET params to the `params` variable.

For `request_params` and `post_params` we may need to change some logic. But this cannot be merged with this change. Probably tests should be changed to fit the new behavior, but this implementation cannot be released with this change.

If you want to investigate what's happening in those you are welcome, if you don't a free slot to do that just notice us and we'll investigate it.

Thanks!

elight:
@jondeandres Seems that the exclusion of `CONTENT_TYPE` causes the `post_params` to come through as `{"{\"key\":\"value\"}"=>nil}`. Updated accordingly to handle when the `CONTENT_TYPE` is absent and `ACCEPT` is a JSON type. Sorry for the awkward back and forth before!

lib/rollbar/request_data_extractor.rb
```diff
@@ -130,6 +136,21 @@ def rollbar_raw_body_params(rack_req)
       rack_req.body.rewind
     end
 
+    def json_request?(rack_req)
+      rack_req.env['CONTENT_TYPE'] =~ %r{application/json} ||
+      rack_req.env['ACCEPT']       =~ /\bjson\b/
+    end
+
+    def json_request_missing_content_type?(env)
```
elight: This is a little clumsier than I would like–but the added work on L1456-147 is necessary to abstract this method.

elight:
Looks like the tests still pass if I merge the `raw_body_params` into the `get_params`. Is this adequate?

As @jingweno pointed out, it's legal for JSON requests to include an `Accept` and not a `ContentType`. As far as I can tell, this change is necessary to support that.

jondeandres:
@elight thanks, let me play with this branch and I'll notify about it ok?

Thanks again!

elight:
Sure thing!

PR #325, opened 2015-10-20
Make scrubbing length configurable

jondeandres:
`Rollbar.configuration.randomize_scrub_length = false` will make us
using `'*' * value.length` as scrubbing value.

When `randomize_scrub_length` option is true, by default, we use a
random string between 3 and 8 chars.

Note: randomize code is duplicated in URL scrubber and the request extractor module. Once we refactor that module we can unify that logic.


coryvirok:
:+1:

PR #331, opened 2015-10-29
implement optional user_ip obfuscator

hqm42:


lib/rollbar/util/obfuscator.rb
```diff
@@ -0,0 +1,15 @@
+module Rollbar
+  module Util
+    module Obfuscator
+      require 'ipaddr'
+      require 'digest'
+
+      def self.obfuscate_ip(ip_string, secret)
+        ip_int32 = IPAddr.new(ip_string, Socket::AF_INET).to_i
```
jondeandres: I'd prefer call this method always and just do `return ip_string unless Rollbar.configuration.user_ip_obfuscator_secret`.

Probably this method should use `rescue` at some point and probably return original ip address string.

lib/rollbar/util/obfuscator.rb
```diff
@@ -0,0 +1,15 @@
...
+        secret_int32 = Digest::MD5.hexdigest(secret)[0..7].to_i(16)
+        obfuscated_ip_int32 = ip_int32 ^ secret_int32 % (2 << 31)
+        IPAddr.new(obfuscated_ip_int32, Socket::AF_INET).to_s
```
jondeandres: Just styling, an space between assignments and returned value would be appreciated :smile:

spec/controllers/home_controller_spec.rb
```diff
@@ -143,6 +143,21 @@
...
+
+      context "rollbar_user_ip obfuscator" do
+        before do
+          Rollbar.configure do |config|
+            config.user_ip_obfuscator_secret = 'secret'
+          end
+        end
+
+        it "should obfuscate the ip when user_ip_obfuscator_secret is set" do
+          real_ip = '1.1.1.1'
```
jondeandres: Thanks for the tests. It'd be nice having more tests with invalid IPs values, so we ensure we don't crash at any point.

jondeandres:
Thanks for the PR @hqm42, I wrote some comments on the code.

hqm42:
@jondeandres: I have added an additional test and refactored my code according to your comments. Rereview please :)

hqm42:
@jondeandres here you go

jondeandres:
Thanks @hqm42, we'll merge this soon.

Issue #237, opened 2015-05-06
custom_data_method controller context


bhh:
maybe its just me but how do u guys use the custom_data_method because it seems i dont have any controller access. in which context is custom_data_method called? i only can find the notifier class.

maybe i use it awfully wrong but atm it seems it can only be filled with static data. cant find any dynamic examples

jondeandres:
@bhh you are right. that method doesn't receive actually any request or context. it can however useful if you need some dynamic data from the host or something.

we have in our roadmap improve that so the defined method receives the current scope, where you will be able to get the request or other data.

Thanks for remind us this :smile:

fabn:
@jondeandres additionally to scope it would be nice to have access (and it should be easier to do) to current exception. For instance I'm dealing with `RestClient::UnprocessableEntity` exceptions and the actual error message is attached to body. Currently I'm sending the error with this code but it's pretty ugly and not thread safe:

```ruby
  config.custom_data_method = lambda do
    if RestClient::Exception === $!
      {
          status_code: $!.http_code,
          errors: (JSON.parse($!.http_body) rescue $!.http_body),
      }
    end
  end
```

jondeandres:
Good point @fab, I agree with you. We'll notice you when we progress on this.

Thanks both!

fab:
Wrong @fab. I think you meant @fabn

jondeandres:
Yah, sorry @fab :smile:

fab:
No worries. It happens every now and then. Probably also because I've contributed to this gem in the past :)

saulcosta18:
Has work been done on this, or do you have a few files you can point me to so I can get started with it?

jondeandres:
@saulcosta18 we know having access to the request in order to send custom data is a very requested and we have plans to do this soon.

The idea would be set a new key in the scope generated by the middleware, for ex in `rollbar-gem/lib/rollbar/middleware/rails/rollbar.rb` and evaluate that value, that should be a `Proc` in `lib/rollbar.rb`, on `#build_payload` method.

We plan to do as soon as we can. Sorry for the inconveniences.

ajsharp:
Any movement on this? It would be good if we could just pass a symbol of a controller method name that would get called.

ArturMoczulski:
@ajsharp I am working on implementing this right now. Should have a pull request ready sometime this week.

siav92:
[...]

PR #341, opened 2015-11-17
Allow access to controller variables from custom_data_method

guanglunw:
Related to #237 

When logging custom data, we'd like access to controller instance variables. This change lets the lambda be evaluated in the context of the controller, which gives access to its variables and methods.

Additionally, it's often useful to specify custom data at the controller level, rather than via a global initialiser. So there's support for this via the following proposed syntax:

``` ruby
include Rollbar::Rails
on_exception_log_values :@instance_var, :any_method
```

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -370,7 +370,7 @@ module Rollbar
     end
 
     def custom_data_method?
-      !!(configuration.custom_data_method || configuration.custom_values)
+      !!(Rollbar.configuration.custom_data_method || Rollbar.configuration.custom_values.count > 0)
     end
 
     def report_custom_data_error(e)
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -231,8 +231,9 @@ module Rollbar
     end
 
     def custom_data
-      data = configuration.custom_data_method.call
+      data = configuration.payload_options[:extra].call
       Rollbar::Util.deep_copy(data)
+
     rescue => e
       return {} if configuration.safely?
 
@@ -344,6 +345,7 @@ module Rollbar
       # Our API doesn't allow null context values, so just delete
       # the key if value is nil.
       data.delete(:context) unless data[:context]
+      data.delete(:extra)
 
       payload = {
         'access_token' => configuration.access_token,
@@ -356,7 +358,9 @@ module Rollbar
     end
 
     def build_payload_body(message, exception, extra)
-      extra = Rollbar::Util.deep_merge(custom_data, extra || {}) if custom_data_method?
+      if custom_data_method? && configuration.payload_options[:extra].respond_to?(:call)
+        extra = Rollbar::Util.deep_merge(custom_data, extra || {})
+      end
 
       if exception
         build_payload_body_exception(message, exception, extra)
@@ -366,7 +370,7 @@ module Rollbar
     end
 
     def custom_data_method?
-      !!configuration.custom_data_method
+      !!(configuration.custom_data_method || configuration.custom_values)
     end
 
     def report_custom_data_error(e)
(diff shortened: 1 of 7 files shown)


jondeandres:
Hi @guanglunw,

thanks for the PR. This is a very requested feature, so yes, thanks again for working on it!

I've liked the idea of having a class method on controllers in order to store which attribute or methods add to the extra data. However I think the code can be simpler if we store that data in the controller itself instead of the configuration, so the code in `RequestDataExtractor` can just ask the controller for the data.

Probably we'll want to take fork this branch, play with it and maybe do some checks since there are quite important changes here.

Beside this, probably more tests on `specs/controllers` would be nice.

I will also do some comments in the code.

lib/rollbar.rb
```diff
@@ -231,8 +231,9 @@ def process_from_async_handler(payload)
     end
 
     def custom_data
-      data = configuration.custom_data_method.call
+      data = configuration.payload_options[:extra].call
```
jondeandres: Probably we are breaking here the public interface for `Configuration`.

lib/rollbar/rails.rb
```diff
@@ -1,5 +1,15 @@
 module Rollbar
   module Rails
 
+    def self.included(base)
+      base.extend ClassMethods
+    end
+
+    module ClassMethods
+      def on_exception_log_values(*vars)
+        Rollbar.configuration.custom_values.merge!({ self.name.parameterize => [*vars].map(&:to_s) })
```
jondeandres: I'd prefer store this data in the controller itself. Another controller method can return the custom data for it according to stored `vars`.

lib/rollbar/request_data_extractor.rb
```diff
@@ -57,6 +57,26 @@ def extract_request_data_from_rack(env)
       data
     end
 
+    def extract_custom_data_from_rack(env)
+      controller = env["action_controller.instance"]
+      if !controller
+        Rollbar.configuration.custom_data_method.call
```
jondeandres: So here we have a `Rack` app and not a `Rails` app, probably we'll want to pass `env` to `#call`?

lib/rollbar/request_data_extractor.rb
```diff
@@ -57,6 +57,26 @@ def extract_request_data_from_rack(env)
...
+        return
+      end
+
+      # Only extract custom data if we've defined a custom_data_method, or specific values on the controller
+      controller_values = Rollbar.configuration.custom_values[controller.class.name.parameterize] || []
+      custom_data_block = Rollbar.configuration.custom_data_method
+      return {} if controller_values.empty? && !custom_data_block
+
+      custom_data = custom_data_block ? controller.instance_exec(&custom_data_block) : {}
```
jondeandres: We can move this logic to some controller method so the controllers are responsible to return the correct logging data. This can be then simpler.

guanglunw:
@jondeandres, you're welcome :-)

Sorry about the lack of tests - ran out of time, but have tested this quite thoroughly against our Rails app and got it to work for our needs...

Feel free to fork and fix - and I do appreciate the comments. They make sense, and as and when I get time in the next few days I'll be happy to revisit this!

jondeandres:
awesome @guanglunw, we'll also try to work on this soon.

we were waiting for a refactor to do on the gem, since I'd like having the extractor of rails apps and rack apps independent. now it's everything like a mix and it's complicating things.

But anyway, we will try to review and check this branch soon.

Thanks!

coryvirok:
@jondeandres what are the next steps for this?

ArturMoczulski:
I submitted a PR for this here: https://github.com/rollbar/rollbar-gem/pull/681

Let me know if there are any features missing. For now I'm closing the issue.

Issue #345, opened 2015-11-23
NoMethodError: undefined method `<' for nil:NilClass in Rollbar::Sidekiq 


aselder:
We're seeing errors in production with Rollbar and Sidekiq. It looks like there is an error in  lib/rollbar/sidekiq.rb:17

The line reads 
`msg_or_context.is_a?(Hash) && msg_or_context["retry"] && msg_or_context["retry_count"] < ::Rollbar.configuration.sidekiq_threshold`

Obviously msg_or_context["retry_count"] is returning nil. It looks like the previous clause in the conditional is trying to guard this, but it looking at a different key.

aselder:
PR to fix this attached

PR #346, opened 2015-11-23
Guard against the retry key being set and no retry_count present

aselder:
This fixes https://github.com/rollbar/rollbar-gem/issues/345

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -14,7 +14,8 @@ module Rollbar
     end
 
     def self.skip_report?(msg_or_context, e)
-      msg_or_context.is_a?(Hash) && msg_or_context["retry"] && msg_or_context["retry_count"] < ::Rollbar.configuration.sidekiq_threshold
+      msg_or_context.is_a?(Hash) && msg_or_context["retry"] && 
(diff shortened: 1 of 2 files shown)


jondeandres:
@aselder which version of sidekiq are you using?

when `msg_or_context['retry_count']` is `nil`, which value of `msg_or_context["retry"]` do you have?

aselder:
@jondeandres We're currently on  sidekiq 2.17.3. We're planning on upgrading soon, but this is a blocker on us using rollbar until it's fixed.

Not sure what the value of  msg_or_context["retry"] is, but I'll try to dig it up with some debuggers.

aselder:
@jondeandres msg_or_context["retry"] is set to 'true' for us, but no retry_count

pacoguzman:
Thanks @aselder I think this is more reliable than the existing version so LGTM. We've checked `retry` because you can configure your worker to not be retried so in that case `retry` has a false value and we don't want to skip the report in that case. And for retriable workers we'll check the retry_count but as you've found we need to be sure we have a value for that key

Good catch thank you!

jondeandres:
@aselder @pacoguzman we've release 2.6.2 with this fix. Thanks.

PR #348, opened 2015-11-25
Add default scrub fields

Crisfole:
Per request from support queue, this adds "api_key" and "access_token" to default scrubbed fields.


ryana:
Rad. I had to have a user roll an API key because it popped up in a Rollbar email. I had seen filtered params before in the rollbar emails, so I incorrectly assumed that api_key and access_token were already in that list. Glad this is up for consideration.

jondeandres:
:+1:

brianr:
:+1:

PR #350, opened 2015-12-01
Use the new Delayed::Job plugin system to register Rollbar plugin

kroky:
Problem: previous Rollbar DelayedJob plugin used Worker Lifecycle
directly to register the around hook. DelayedJob now clears and
reinitializes the Lifecycle object upon Worker initializations.
This means that commands like `rake jobs:work` will first register
Rollbar hook, then initialize Worker which will clear the hook.

Solution: use the new plugins system to allow Worker to setup its
Lifecycle object and register all plugins when it needs to.

### DIFF
--- a/lib/rollbar/delayed_job.rb
+++ b/lib/rollbar/delayed_job.rb
@@ -23,12 +23,18 @@ module Rollbar
       end
     end
 
+    class RollbarPlugin < ::Delayed::Plugin
+      callbacks do |lifecycle|
+        lifecycle.around(:invoke_job, &Delayed::invoke_job_callback)
+      end
+    end
+
     self.wrapped = false
 
     def self.wrap_worker
       return if wrapped
 
-      around_invoke_job(&invoke_job_callback)
+      ::Delayed::Worker.plugins << RollbarPlugin
 
       self.wrapped = true
     end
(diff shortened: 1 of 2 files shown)


jondeandres:
Hi @kroky,

thanks for this PR. It's a nice change :smile:. Do you know for which versions of delayed job this is supported? We'll need to use different approaches for old versions?

Thanks for this!

kroky:
Hi @jondeandres,

I just checked 4.0.0 and 3.0.0 and they both had the Plugins system. Not sure if you support v2 but it seems a complete rewrite from that time and it is back in 2010. It is probably wise to test this change with versions back to 3.0.0.

jondeandres:
awesome @kroky, we will test and review this and probably merge it if everything is working fine.

thanks a lot for this, again, it's a nice change!

kroky:
:+1: thanks! I stumbled upon this today and decided to send you the update as we are using Rollbar with delayed job bg workers and having exceptions reported is important. Thanks for all your work with Rollbar!

jondeandres:
@kroky from your PR description:

```DelayedJob now clears and
reinitializes the Lifecycle object upon Worker initializations.
This means that commands like rake jobs:work will first register
Rollbar hook, then initialize Worker which will clear the hook.```

From which version have you detected this?

Thanks

kroky:
Well, I am using v4.1.1 and it fails with the old code. I am not sure when did they introduce the change which is this:
https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/worker.rb#L136

jondeandres:
Nice, seems they added that change from 4.1.0.

Thanks!

Issue #330, opened 2015-10-28
build_payload triggers URI::InvalidURIError


fabn:
Getting a lot of these errors in the wild

```
Failsafe from rollbar-gem. URI::InvalidURIError in .../1.9.1/uri/common.rb:176:in `split': build_payload in exception_data
```

I think the cause might be [this line](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/scrubbers/url.rb#L24) but I don't have any clue on how to reproduce this because request data are not reported in rollbar and I can't find them in any log.

This is very annoying and I want to prevent this error to remove some noise from my RB account.

jondeandres:
@fabn any idea about the URL you are trying to parse?

This is Ruby 1.9.1?

fabn:
> @fabn any idea about the URL you are trying to parse?

No, it's a report from a high traffic app, I get this few times per day. I think that is caused by some misbehaving bot that sends out bad UTF-8 data in some header or query param.
 
> This is Ruby 1.9.1?

No, current code is running on latest 1.9.3 MRI that 1.9.1 is part of library path created by rvm, full path is this
```
/usr/local/rvm/rubies/ruby-1.9.3-p551/lib/ruby/1.9.1/uri/common.rb:176
```

This is specific to my app however I think the point here is that you should provide the option to save a full dump of the report data somewhere in the filesystem in [these lines](https://github.com/rollbar/rollbar-gem/blob/574b14717aa37728c22e1e66c600cb3e119280e0/lib/rollbar.rb#L143-L149) in order to diagnose this kind of issues.

jondeandres:
@fab thanks for the info. we will rescue from `URI::InvalidURIError` which is probably the best idea since  `URI` behavior changes with different Ruby versions. About logging the dump, you are totally right. We'll try to improve this.

I'll close this issue once we merge a fix for this.

Thanks Again

fabn:
Another issue I'm getting (partially related to this issue) is this

```
Failsafe from rollbar-gem. NoMethodError in rollbar-2.5.1/lib/rollbar/util.rb:69:in `block in deep_merge': build_payload in exception_data
```

Also in this report I'm completely losing context of the error and it's hard to diagnose.

In my opinion you should really add exception message in [this code](https://github.com/rollbar/rollbar-gem/blob/574b14717aa37728c22e1e66c600cb3e119280e0/lib/rollbar.rb#L626-L633), in this way at least some info are preserved in error messages. 

[...]

PR #354, opened 2015-12-07
Better diagnostic when failsafe is on

fabn:
First commit is about [this comment](https://github.com/rollbar/rollbar-gem/issues/330#issuecomment-152162956) there's no reason to swallow exception message, so add it to report.

Second commit is about an issue I have in reports with your code. I don't know where that method is used, but I get this error in my reports

```
Failsafe from rollbar-gem. NoMethodError in [...]/rollbar-2.6.3/lib/rollbar/util.rb:70:in `block in deep_merge': build_payload in exception_data
```

Looking at [current implementation](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/util.rb#L70) the only possible reason for this message is that `hash1[k]` is nil for some `k` (since you're iterating on `hash2` keys). The same issue could happen two line below when doing the same check for `Array`.

I'd like to get this merged because this error could potentially swallow other application errors.

Thanks.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -618,6 +618,8 @@ module Rollbar
           nearest_frame = backtrace[0]
 
           exception_info = exception.class.name
+          # #to_s and #message defaults to class.to_s. Add message only if add valuable info.
+          exception_info += %Q{: "#{exception.message}"} if exception.message != exception.class.to_s
           exception_info += " in #{nearest_frame}" if nearest_frame
 
           body += "#{exception_info}: #{message}"
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -1575,6 +1575,17 @@ describe Rollbar do
       notifier.send(:send_failsafe, nil, nil)
     end
 
+    context 'with a non default exception message' do
+      let(:exception) { StandardError.new 'Something is wrong' }
+
+      it 'adds it to exception info' do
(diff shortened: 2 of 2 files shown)


fabn:
Added a test for the exception message and fixed the used operator (lesson to myself: never submit a PR without running the code).

lib/rollbar/util.rb
```diff
@@ -66,9 +66,9 @@ def self.deep_copy(obj)
 
     def self.deep_merge(hash1, hash2)
       hash2.each_key do |k|
-        if hash1[k].is_a? ::Hash and hash2[k].is_a? ::Hash
+        if hash1[k] and hash1[k].is_a? ::Hash and hash2[k].is_a? ::Hash
```
jondeandres: If we are actually checking that `hash1[k]` is a `Hash` object, why we need to check it exists? 

`nil.is_a?(Hash) # false`

fabn:
Right, `NilClass` responds to `#is_a?` so my error should be caused by something else, or maybe the backtrace line reported is wrong.

Would you accept a rebased PR with only the first commit and the test? In this way we'd have more context for these errors.

jondeandres:
@fabn sure! we'll be glad to merge this PR with the first message fix commit.

fabn:
Rebased and squashed.

fabn:
@jondeandres rebased and squashed

PR #357, opened 2015-12-10
Ignore exception cause when it's not another exception

pedro:
Custom error classes might define `#cause` as a string or similar, which would in turn result in an internal Rollbar error.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -442,7 +442,7 @@ module Rollbar
       traces = [trace_data(exception)]
       visited = [exception]
 
-      while exception.respond_to?(:cause) && (cause = exception.cause) && !visited.include?(cause)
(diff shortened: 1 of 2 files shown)


jondeandres:
@pedro do you have this error using third party libraries? the problem is with some custom exceptions of your own?

I remember we had a problem with another library that defined its own `#cause` method, but upgrading the library resolved the problem.

Can you give us those details?

Thanks

pedro:
Hey! For sure, it is actually a private library, so we'll probably end up changing the method name.

But seeing that Ruby didn't give us any warning when we defined that method, and that debugging this in Rollbar took quite some time, I figured it might be worth being extra cautious here. [Searching in Github this method doesn't seem all uncommon](https://github.com/search?l=ruby&p=7&q=%22def+cause%22&ref=searchresults&type=Code&utf8=%E2%9C%93) (wish there was a way to skip duplicates heh).

PR #362, opened 2015-12-17
Fix Rollbar::Util.deep_merge so it has default values for arguments

### LINKED ISSUE #358: nil value for Rollbar::Util.deep_merge first argument
With release 2.7.0 I got the actual error message for the issued described in #354 that is

```
Failsafe from rollbar-gem. NoMethodError: "undefined method `[]' for nil:NilClass" in lib/rollbar/util.rb:70:in `block in deep_merge': build_payload in exception_data
```

So instead of the patch discussed [here](https://github.com/rollbar/rollbar-gem/pull/354#discussion-diff-46841517L69) the right fix should check if `hash1` is nil.

However I don't know how to handle that case because I don't know the semantic of that method.

Looking at the code a simple `hash1 ||= {}` before the `each_key` loop may fix the issue but I don't know what consequences may have on the other code, so I won't submit a PR for that.

jondeandres:
We want the default value for the arguments in
`Rollbar::Util.deep_merge` to be empty Hashes.

Fixes #358

### DIFF
--- a/lib/rollbar/util.rb
+++ b/lib/rollbar/util.rb
@@ -65,11 +65,11 @@ module Rollbar
       end
     end
 
-    def self.deep_merge(hash1, hash2)
+    def self.deep_merge(hash1 = {}, hash2 = {})
       hash2.each_key do |k|
-        if hash1[k].is_a? ::Hash and hash2[k].is_a? ::Hash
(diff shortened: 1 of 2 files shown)


brianr:
Can you add a test for the bug this fixes?

jondeandres:
Sure, I pushed a test and a change in the implementation. I'll merge and release.

PR #368, opened 2016-01-28
Notify deploy only on primary :rollbar_role server

theist:
Hi, I've seen that on every deploy on a multi app node the deploy gets noticed once for each node which get the deploy, so on a tree app node connected to rollbar there's three notifications of the same deploy which makes tree notices on a slack channel if you have slack and rollbar paired.

So perhaps it's a good idea notify the deploy only on primary server.

*Important note*: I've made this PR using the github editor and my knowledge of capistrano so it is not properly tested. I planed only to open an issue but the change is so small that I've created it as PR.

### DIFF
--- a/lib/rollbar/tasks/rollbar.cap
+++ b/lib/rollbar/tasks/rollbar.cap
@@ -6,7 +6,7 @@ namespace :rollbar do
 
   desc 'Send the deployment notification to Rollbar.'
   task :deploy do
-    on roles fetch(:rollbar_role) do
+    on primary fetch(:rollbar_role) do
       warn("You need to upgrade capistrano to '>= 3.1' version in order to correctly report deploys to Rollbar. (On 3.0, the reported revision will be incorrect.)") if Capistrano::VERSION =~ /^3\.0/
 
       uri    = URI.parse 'https://api.rollbar.com/api/1/deploy/'


jondeandres:
Hi @theist,

I think the PR is ok, but I'll need to check the compatibility with different versions of capistrano.

Have you know the behavior if the user has only one server and didn't flag the server as `primary`? Maybe the task will not be executed?

Thanks

theist:
I'm used to the last version of capistrano. On 3.4 the first server defined for a role is automagically defined as primary so primary :role always return a host. But capistrano changes fast. Perhaps at weekend I be able to do some homework, and check properly this PR

theist:
Hi @jondeandres 

I've checked capistrano 3.0 and 3.1 behaviour about roles/primary

As in 3.4, in capistrano 3.0 the first server for a defined role is automatically considered primary. So there's always a primary and only one server for each role considered "primary"

This is the current template of "stage" and the text is similar in 3.1 https://github.com/capistrano/capistrano/blob/master/lib/capistrano/templates/stage.rb.erb#L15

I think that there's the code which selects primary : 
https://github.com/capistrano/capistrano/blob/master/lib/capistrano/configuration/servers.rb#L50

I also tested capistrano 3.0/3.1 with this config 

```
server '127.1.1.1', user: 'test', roles: %w{web app}
server '127.2.2.2', user: 'test', roles: %w{web app}
server '127.3.3.3', user: 'test', roles: %w{web app}

require "pry"
task :debug do
  binding.pry
end
```

And did these tests with pry (this is cap 3.0)


```
[3] pry(main)> on roles(:app) do |host| 
[3] pry(main)*   puts host.hostname 
[3] pry(main)* end  
127.2.2.2
127.3.3.3
127.1.1.1
=> [#<Thread:0x007f25da660e98@/home/carlos/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/sshkit-1.7.1/lib/sshkit/runners/parallel.rb:11 dead>,
 #<Thread:0x007f25da660d58@/home/carlos/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/sshkit-1.7.1/lib/sshkit/runners/parallel.rb:11 dead>,
 #<Thread:0x007f25da660c40@/home/carlos/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/sshkit-1.7.1/lib/sshkit/runners/parallel.rb:11 dead>]

[4] pry(main)> primary(:app)
=> #<Capistrano::Configuration::Server:0x007fb6610615c0
 @hostname="127.1.1.1",
 @port=nil,
 @properties=
  #<Capistrano::Configuration::Server::Properties:0x007fb661057ca0
   @properties={},
   @roles=#<Set: {:web, :app}>>,
 @user="carlos">

[5] pry(main)> primary(:app).hostname
=> "127.1.1.1"

[6] pry(main)> on primary(:app) do |host|
[6] pry(main)*   puts host.hostname
[6] pry(main)* end  
=> [#<Thread:0x007fb66341f250@/home/carlos/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/sshkit-1.7.1/lib/sshkit/runners/parallel.rb:11 dead>]

[7] pry(main)> 
```

jondeandres:
Awesome @theist! Thanks for debugging this and resolve the doubts. I'll merge this today, thanks for this!

jondeandres:
@theist could you sync the branch with master? We fixed a problem with the sucker punch version and your tests are failing cause that.

Thanks!

theist:
:P sorry did accidentally a two way merge while triying to sync repos

theist:
Took two tries but seems ok now. Sorry about the double CI cycle. Now its pushed only my change against the current master.

jondeandres:
awesome @theist, thanks!

PR #373, opened 2016-02-02
[WIP] Add rough draft of configuration options documentation

Crisfole:
@jondeandres @brianr 

This is a first pass at each one. I stuck a few `TODO` where I was extremely uncertain, and it's not organized the way I want yet (see the TODO at the top).

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -16,8 +16,8 @@ and respect the rate limits set on that access token.
 ## async_handler
 
 If `config.use_async = true` explicitly sets the function used to send
-asynchronous payloads to Rollbar. Not needed if using one of the built in async
-reporters:
+asynchronous payloads to Rollbar. Should be an object that responds to `call`
+Not needed if using one of the built in async reporters:
 
  * girl_friday
  * sucker_punch
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -43,7 +43,21 @@ Rollbar understands these formats:
 
 ## custom_data_method
 
-TODO: What does it do? Demo?
+The method to call to gather custom data to send with each rollbar request.
+
+```ruby
+def custom_data
+  return {
+    :custom => {
+      :key1 => get_key_one,
+      :key2 => get_key_two
+    },
+    :server => {
+      :root => '/home/username/www/'
+    }
+  }
+end
+```
 
 ## delayed_job_enabled
 
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -27,8 +27,6 @@ Not needed if using one of the built in async reporters:
 
 ### branch
 
-**Default** "master"
-
 Name of the checked-out source control branch.
 
 ### code_version
@@ -70,10 +68,10 @@ operation.
 
 **Default** `false`
 
-Disables monkey patching all non-core monkey patches. If you do this you'll
-need to manually `use` an appropriate Rollbar middleware.
+Disables monkey patching all non-core monkey patches and automatic reporting.
 
-Especially useful if you're not using Rails or Sinatra.
+If you set this to true you will be responsible for rescuing and reporting all
+errors manually.
 
 ### disable_core_monkey_patch
 
@@ -89,16 +87,15 @@ Be careful using this option as it may caused unexpected behavior in some situat
 **Default** `true`
 
 
-Set to false if you have `Delayed`  but do not wish to wrap Delayed jobs with a
+Set to false if you have `delayed_job`  but do not wish to wrap jobs with a
 Rollbar notifier.
 
 ### report_dj_data
 
 **Default** `true`
 
-Set to `false` to skip automatic reporting of Delayed job data. This can be
-handy if you manually report to Rollbar, or you have another way of catching and
-reporting those errors.
+Set to `false` to skip automatic bundling of job metadata like queue, job class
+name, and job options.
 
 ### dj_threshold
 
@@ -108,8 +105,8 @@ The number of job failures before reporting the failure to Rollbar.
 
 ### failover_handlers
 
-An array of backup handlers if the async handlers fail. Each should respond to
-`call` and should receive a `payload`.
+An array of backup handlers if the async handlers fails. Each should respond to
+`#call` and should receive a `payload`.
 
 ### filepath
 
@@ -132,8 +129,8 @@ properly defined `person_method` or `person_id_method`.
 
 ### logger
 
-The logger to use *instead of* the default logger. Especially useful in `scope`s
-where you wish to send log messages elsewhere.
+The logger to use *instead of* the default logger. Especially useful when you
+wish to send log messages elsewhere.
 
 ### payload_options
 
@@ -143,36 +140,39 @@ Extra data to send with the payload.
 
 Rails only: A string or symbol giving the name of the method on the controller.
 Should return an object with an `id` method, and optionally `username` and
-`email` methods.
+`email` methods. The names of the `id`, `username` and `email` methods can be
+overridden. See `person_id_method`, `person_username_method`, and
+`person_email_method`.
 
 If not using Rails:
 
-Populate the Rack `person_data` key with a hash containing `:id`, and optionally
-`:username` and `:email`.
+Populate the `rollbar.person_data` key with a hash containing `:id`, and
+optionally `:username` and `:email`.
 
 ### person_id_method
 
-Rails only: a string or symbol giving the name of the method on the controller
-which returns the current user's id (a string). Ignored if `person_method`
-present.
+A string or symbol giving the name of the method on the user instance that
+returns the person's id. Gets called on the result of `person_method`. Ignored
+if `person_method` not present.
 
 ### person_username_method
 
-Rails only: a string or symbol giving the name of the method on the controller
-which returns the current user's username. Ignored if `person_id_method` not
-present. Ignored if `person_method` present.
+A string or symbol giving the name of the method on the user instance that
+returns the person's username. Gets called on the result of `person_method`.
+Ignored if `person_method` not present.
 
 ### person_email_method
 
-Rails only: a string or symbol giving the name of the method on the controller
-which returns the current user's email address. Ignored if `person_id_method`
-not present. Ignored if `person_method` present.
+A string or symbol giving the name of the method on the user instance that
+returns the person's email. Gets called on the result of `person_method`.
+Ignored if `person_method` not present.
 
 ### populate_empty_backtraces
 
-If you report a `new` exception, but do not `raise` it, the backtraces in Rollbar
-will be empty. Set `populate_empty_backtraces` to `true` to have Rollbar load
-the traces before sending them.
+Raising an exception in Ruby is what populates the backtraces. If you report a
+manually initialized exception instead of a raised and rescued exception, the
+backtraces will be empty. Set `populate_empty_backtraces` to `true` to have
+Rollbar load the traces before sending them.
 
 ### request_timeout
 
@@ -251,7 +251,7 @@ SSL connection problems.
 **Default** `false`
 
 When `true` indicates you wish to send data to Rollbar asynchronously. If
-installed, uses `girl_friday`, otherwise defaults to `Threading`.
+installed, uses `girl_friday`, otherwise defaults to `Thread`.
 
 ### use_eventmachine
 
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -1,0 +1,303 @@
+# Configuration Reference
+
+TODO: SECTIONS(DELAYED JOB SUPPORT, ASYNC SUPPORT, WRITE TO FILE, PERSON TRACKING)
+
+<!-- Sub:[TOC] -->
+
+## access_token
+
+**Required**
+
+Sets the access token used to send payloads to rollbar.
+
+Items sent through a given access token arrive in that access token's project
+and respect the rate limits set on that access token.
+
+## async_handler
+
+If `config.use_async = true` explicitly sets the function used to send
+asynchronous payloads to Rollbar. Not needed if using one of the built in async
+reporters:
+
+ * girl_friday
+ * sucker_punch
+ * Sidekiq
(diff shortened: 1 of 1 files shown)


docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+**Required**
+
+Sets the access token used to send payloads to rollbar.
+
+Items sent through a given access token arrive in that access token's project
+and respect the rate limits set on that access token.
+
+## async_handler
+
+If `config.use_async = true` explicitly sets the function used to send
```
jondeandres: The object passed to this option should respond to `#call` and receive the payload. Examples in `lib/rollbar/delay/`.

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+
+Rollbar understands these formats:
+
+ * semantic version (i.e. "2.1.12")
+ * integer (i.e. "45")
+ * git SHA (i.e. "3da541559918a808c2402bba5012f6c60b27661c")
+
+## custom_data_method
+
+TODO: What does it do? Demo?
```
jondeandres: It's just a method called to add more info to the payload.


```ruby
def custom_data
      data = configuration.custom_data_method.call
      Rollbar::Util.deep_copy(data)
    rescue => e
      return {} if configuration.safely?

      report_custom_data_error(e)
    end
```

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+**Default** `Logger.new(STDERR)`
+
+What logger to use for printing debugging and informational messages during
+operation.
+
+## disable_monkey_patch
+
+**Default** `false`
+
+Disables monkey patching `Rack::Builder`. If you do this you'll need to manually
```
jondeandres: Not only for `Rack::Builder`, also for any other monkey patch we have.

```ruby
def require_hooks
      return if configuration.disable_monkey_patch
      wrap_delayed_worker

      require 'rollbar/active_record_extension' if defined?(ActiveRecord)
      require 'rollbar/sidekiq' if defined?(Sidekiq)
      require 'rollbar/active_job' if defined?(ActiveJob)
      require 'rollbar/goalie' if defined?(Goalie)
      require 'rollbar/rack' if defined?(Rack)
      require 'rollbar/rake' if defined?(Rake)
    end
```

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+
+Especially useful if you're not using Rails or Sinatra.
+
+## disable_core_monkey_patch
+
+**Default** `false`
+
+TODO: VERIFY
+
+Disables monkey patching certain core features without which Rollbar may not
```
jondeandres: For now this disables the basic socket monkey patch if active support is defined. Rollbar can work without it except in some specific cases.

There is another monkey patch, to Thread, that cannot be removed and it's always applied.

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+properly defined `person_method` or `person_id_method`.
+
+## logger
+
+The logger to use *instead of* the default logger. Especially useful in `scope`s
+where you wish to send log messages elsewhere.
+
+## payload_options
+
+Extra data to send with the payload. Use `scope` with `payload_options` to set
```
jondeandres: I'd remove the `scope` reference here.

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+## root
+
+Set the server root, all stack frames outside that root are considered
+'non-project' frames. Also used to setup Github linking.
+
+## safely
+
+**Default** `false`
+
+When `true` evaluates `custom_data_method` in a `begin`/`rescue` block, returning
```
jondeandres: We always evaluate the method in a `begin/rescue` block, but if there is an exception there and we are in `safely` mode, we just return `{}` there.

jondeandres:
Nice work @Crisfole :smile:. I just wrote some inline comments in the docs.

docs/configuration_options.md
Crisfole: `deep_copy` knows where to stick the extra data?

docs/configuration_options.md
Crisfole: So....how shoudl we document this? (Shoudl we leave it undocumented for people who know what they're doing?) Why'd we add it?

docs/configuration_options.md
```diff
@@ -0,0 +1,303 @@
...
+**Default** `true`
+
+TODO: CORRECT THIS
+
+When true filters the user out of the URL.
+
+## scrub_password
+
+TODO: CORRECT THIS
+
```
Crisfole: @jondeandres ???

docs/configuration_options.md
jondeandres: @Crisfole deep_copy call is there in order to be sure we don't mutate never the value received from the custom_data_method.

So, we call `configuration.custom_data_method.call` and then we do a deep copy with the result, so it's a new object and can't the result object.

docs/configuration_options.md
jondeandres: We should document it yes. Just explaining that this disables our monkey patches in the ruby core libary.

docs/configuration_options.md
jondeandres: These values should be boolean and the purpose is what you have already there.

When scrubbing the URL, we scrub or not the user/password. By default true.

So, scrubbing URLs, we always scrub the query string, and the user and password is configurable. We'll scrub by default those.

Crisfole: Who sends their *password* in the URL!? Or does this scrub access tokens too?

That's not hidden by HTTPS...If they're sending the password in plaintext in the URL they've got bigger problems that us recording it in Rollbar!

Crisfole:
I'm still not 100% certain I have all this right.

@jondeandres 
 * What’s the difference between `delayed_job_enabled` and `report_dj_data`?
 * What about the `person_method*`? Is it really rails only? Or does it accept anything that responds to call?
 * Can someone go over this in detail? I know it's a pain, but it's the only way to make sure we're not missing anything obvious.

jondeandres:
@Crisfole if `delayed_job_enabled` is enabled we'll wrap the delayed_job worker so the errors in the jobs are reported.

Instead of this, there is some delayed_job data to be sent or not. We can disabled it with `report_dj_data = false`. So, the job itself has some metadata related, that is the data to be sent or not.

Crisfole:
@jondeandres what sort of metadata are we talking?

jondeandres:
@Crisfole job class, queue...

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+ * girl_friday
+ * sucker_punch
+ * Sidekiq
+ * Resque
+ * threading
+
+### branch
+
+**Default** "master"
```
jondeandres: Looking at our code I cannot see a default `master` value for this.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+
+What logger to use for printing debugging and informational messages during
+operation.
+
+### disable_monkey_patch
+
+**Default** `false`
+
+Disables monkey patching all non-core monkey patches. If you do this you'll
+need to manually `use` an appropriate Rollbar middleware.
```
jondeandres: Do you think this it's well explained? It's true that `use` is the method used in rack based frameworks but may not be for other libaries/frameworks.

I'd just explain that we disable all the monkey patches and auto-reporting.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+Be careful using this option as it may caused unexpected behavior in some situations.
+
+## Delayed Job Monitoring
+
+### delayed_job_enabled
+
+**Default** `true`
+
+
+Set to false if you have `Delayed`  but do not wish to wrap Delayed jobs with a
```
jondeandres: the name of the library is "delayed_job". and I'd change "wrap Delayed jobs" with just "wrap jobs".

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+Rollbar notifier.
+
+### report_dj_data
+
+**Default** `true`
+
+Set to `false` to skip automatic reporting of Delayed job data. This can be
```
jondeandres: This is not really true.

If `delayed_job_enabled` is true we'll rescue the jobs exceptions and send the backtrace information.

When reporting a job exception, if `report_dj_data` is set to `true`, we'll send the job details information, not only the exception, backtrace, etc...

The job details information is the queue, the job class name, and different jobs options.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+
+### dj_threshold
+
+**Default** 0 (Report *any* errors)
+
+The number of job failures before reporting the failure to Rollbar.
+
+### failover_handlers
+
+An array of backup handlers if the async handlers fail. Each should respond to
```
jondeandres: should be "if the async handler fails" and change `call` with `#call`.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+
+Sets the access token used to send payloads to rollbar.
+
+Items sent through a given access token arrive in that access token's project
+and respect the rate limits set on that access token.
+
+### async_handler
+
+If `config.use_async = true` explicitly sets the function used to send
+asynchronous payloads to Rollbar. Should be an object that responds to `call`
```
jondeandres: `call` => `#call`.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+### ignored_person_ids
+
+**Default** `[]`
+
+Ids of people whose reports you wish to ignore. Only works in conjunction with a
+properly defined `person_method` or `person_id_method`.
+
+### logger
+
+The logger to use *instead of* the default logger. Especially useful in `scope`s
```
jondeandres: I shouldn't use `scope` there but just "scope" or "scenarios".

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+
+### person_method
+
+Rails only: A string or symbol giving the name of the method on the controller.
+Should return an object with an `id` method, and optionally `username` and
+`email` methods.
+
+If not using Rails:
+
+Populate the Rack `person_data` key with a hash containing `:id`, and optionally
```
jondeandres: Customers need to populate the key `rollbar.person_data` not `person_data` in the Rack environment. It's important emphasize the Rack environment.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+`:username` and `:email`.
+
+### person_id_method
+
+Rails only: a string or symbol giving the name of the method on the controller
```
jondeandres: It's not the method on the controller but the method in the user instance found after calling the `person_method`.

```ruby
def rollbar_person_data
        user = send(Rollbar.configuration.person_method)
        # include id, username, email if non-empty
        if user
          {
            :id => (user.send(Rollbar.configuration.person_id_method) rescue nil),
            :username => (user.send(Rollbar.configuration.person_username_method) rescue nil),
            :email => (user.send(Rollbar.configuration.person_email_method) rescue nil)
          }
        else
          {}
        end
      rescue NoMethodError, NameError
        {}
      end
````

File: docs/configuration_options.md [hunk 17]
```diff
@@ -0,0 +1,275 @@
...
+which returns the current user's id (a string). Ignored if `person_method`
+present.
+
+### person_username_method
+
+Rails only: a string or symbol giving the name of the method on the controller
```
jondeandres: Same as above.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+which returns the current user's username. Ignored if `person_id_method` not
+present. Ignored if `person_method` present.
+
+### person_email_method
+
+Rails only: a string or symbol giving the name of the method on the controller
```
jondeandres: Same as above.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+which returns the current user's email address. Ignored if `person_id_method`
+not present. Ignored if `person_method` present.
+
+### populate_empty_backtraces
+
+If you report a `new` exception, but do not `raise` it, the backtraces in Rollbar
```
jondeandres: I'd explain: if you report a manually initialized exception instead of a raised and rescued exception, the backtraces will be empty.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+By default we use `OpenSSL::SSL::VERIFY_PEER` for SSL. Although we don't
+recommend changing it, you can disable peer verification in case you experience
+SSL connection problems.
+
+### use_async
+
+**Default** `false`
+
+When `true` indicates you wish to send data to Rollbar asynchronously. If
+installed, uses `girl_friday`, otherwise defaults to `Threading`.
```
jondeandres: If you use the ` characters you should use existing methods, constants, etc...

In ruby `Threading` is not a constant and doesn't exist in the core library, use `Thread` instead.

docs/configuration_options.md
```diff
@@ -0,0 +1,275 @@
...
+
+### use_eventmachine
+
+**Default** `false`
+
+When `true` indicates you wish to send data to Rollbar with `eventmachine`.
```
jondeandres: `EventMachine` instead of `eventmachine`.

jondeandres: Forget this, you are referring the gem. It's ok.

docs/configuration_options.md
Crisfole: The default is master on the server side.

Crisfole: I'll toss it anyway.

docs/configuration_options.md
Crisfole: Good catch! Will you teach me English? ;)

docs/configuration_options.md
Crisfole: Oops, didn't realize that.

Crisfole:
@jondeandres All set, I believe.

PR #375, opened 2016-02-03
Add before_process and transform hooks

jondeandres:


brianr:
@jondeandres It seems that the `:scope` is empty/missing in the `before_process` hooks. Repro case:

```
handler1 = proc do |options|
  puts "+++ in handler1"
  puts options
  puts options[:scope][:custom][:some_key]
  puts "+++"
end

Rollbar.configure do |config|
  config.access_token = 'token'
  config.before_process << handler1
end

Rollbar.scope(:custom => {:some_key => "some value"}).debug("a debug message")
```

When I do the above, the output is:

```
+++ in handler1
{:level=>"debug", :scope=>{}, :exception=>nil, :message=>"a debug message", :extra=>nil}
E, [2016-02-10T13:42:33.745977 #20753] ERROR -- : [Rollbar] Error calling the `before_process` hook: undefined method `[]' for nil:NilClass
```

jondeandres:
@brianr the interface to access the scope is `options[:scope].custom` in that case. But using `[]` shouldn't give that behavior. I'll review this.

jondeandres:
Forget it, it's a bug yes. I'm passing there `configuration.payload_options` instead of the `scope_object`. I'll fix it.

jondeandres:
@brianr I've fixed and changed the interface in https://github.com/rollbar/rollbar-gem/commit/3428eb13352047fb6c43291e50c9cc8ddb73811c.

I think it'd be better if the users can access the scope as you wanted, `scope[:request]` instead of `scope.request`, since the scope is defined like `Rollbar.scoped(request: some_data)`.

PR #376, opened 2016-02-03
Fix Rollbar::Notifier#report_custom_data_error

jondeandres:
Ensure that we receive a Hash object so we can take the uuid from
there. In any other case just return an empty Hash.

[branch #5660]


brianr:
:+1:

PR #377, opened 2016-02-03
Add disable_rack_monkey_patch configuration option

jondeandres:
[branch #3860]


brianr:
Nice. Should we also add some docs for when this should be used?

jondeandres:
@brianr I think we can merge this and add that doc to https://github.com/rollbar/rollbar-gem/pull/373.

@Crisfole can you do it?

PR #378, opened 2016-02-03
Reset the notifier when we use Rollbar.preconfigure

jondeandres:
This fix the problem we had with gems using `Thread.new` before our
railtie is loaded.

The order to reproduce the bug was:

1. Load Rollbar gem and the Thread monkey patch
2. Rollbar.configure is executed
3. Another gem is loaded and executes Thread.new

  3.1 This sets Rollbar.notifier to the current configuration
4. Our railtie is loaded

 4.1 The railtie uses preconfigure that changes Rollbar.configuration
 4.2 Because we already have a Rollbar.notifier, the new configuration
      is not propagated

So now, on `Rollbar.preconfigure` we reset the notifier, so for next
time it's used it will take the last configuration, included the railtie one.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -744,6 +744,8 @@ module Rollbar
     # to configure it without initializing any of the third party hooks
     def preconfigure
       yield(configuration)
+
+      reset_notifier!
     end
 
     def configure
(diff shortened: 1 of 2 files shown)


brianr:
:+1:

PR #382, opened 2016-02-09
Rollbar js instrumentation

jondeandres:
Users can now use the gem to configure the Rollbar JS notifier using the
same initializer.

The gem will inject the Rollbar JS snippet and the configuration to the
returned response.

In order to configure the Rollbar.js using the gem, users can follow the
next:

```ruby
js_config = {
        accessToken: "POST_CLIENT_ITEM_ACCESS_TOKEN",
        captureUncaught: true,
        payload: {
          environment: "production"
        }
}

Rollbar.configure do |config|
  # common gem configuration
  # ...
  config.js_enabled = true
  config.js_options = js_config
end
```

The options passed to `#js_options=` are the same than the used for the
Rollbar JS notifier.


Crisfole:
:+1: Nice!

PR #386, opened 2016-02-18
Fix exception_level_filter

Crisfole:
Ruby `:` syntax for hashes *only* supports Symbols as keys. Since we expect a string this doesn't work.


magnusvk:
:+1:

jondeandres:
@Crisfole :+1:

PR #387, opened 2016-02-18
Update readme

guillewu:
Updating incorrect syntax in readme.


brianr:
Thanks @guillewu ! We've got the same fix via #386.

PR #392, opened 2016-02-29
Allow Rollbar to work in Rails < 3 in Ruby mode

lleirborras:
Fo those who are still maintaining old Rails apps this is a way to integrate the Rollbar gem as if it was a plain Ruby project giving the ability to execute `Rollbar.error('foo')` manually in the code whenever is needed.

### DIFF
--- a/lib/rollbar.rb
+++ b/lib/rollbar.rb
@@ -17,7 +17,7 @@ require 'rollbar/encoding'
 require 'rollbar/logger_proxy'
 require 'rollbar/exception_reporter'
 require 'rollbar/util'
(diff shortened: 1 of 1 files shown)


brianr:
Thanks @lleirborras !

This seems fine to me - @jondeandres can you review?

jondeandres:
Hi @lleirborras, thanks for the PR!

I think we will merge this PR, but we don't think we'll support officialy Rails < 3.0, but we are happy you'll be able to use the gem!

lleirborras:
Thanks!
I understand Rails < 3 is not on the roadmap but I managed to have my notifications working just fine. Great job guys!

PR #394, opened 2016-03-04
Rollbar JavaScript: Make <head> lookup stricter to avoid matching <header>

jeremiemv:
When loading an HTML partial including a \<header\> tag via AJAX, the Rollbar JS is inserted inside that said \<header\> tag, the lookup for the \<head\> tag being too permissive.

screen shot 2016-03-04 at 11 18 00 am

### DIFF
--- a/lib/rollbar/js/middleware.rb
+++ b/lib/rollbar/js/middleware.rb
@@ -90,7 +90,7 @@ module Rollbar
       end
 
       def find_end_of_head_open(body)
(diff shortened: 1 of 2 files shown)


brianr:
Thanks @jeremiemv ! This looks good to me. @jondeandres can you take a look?

jondeandres:
Hey @jeremiemv, thanks for the PR!

jeremiemv:
:+1:

PR #395, opened 2016-03-05
Fix encoding problem when scrubing parameters

### LINKED ISSUE #389: #145 Failsafe from rollbar-gem. ArgumentError: "invalid byte sequence in UTF-8" in /usr/local/rvm/gems/ruby-2.2.1/gems/rollbar-2.7.1/lib/rollbar/request_data_extractor.rb:197:in `=~': build_payload in exception_data
We are getting this error so much. Once every 20 minute. What is the reason of this error, how can we avoid this?

jondeandres:
Fix #389


brianr:
:+1:

jondeandres:
@brianr it seems that the tests are not passing on ruby <= 2.0, I need to investigate more.

Issue #396, opened 2016-03-09
Incompability with rake (11.0.1): cannot determine rake version


pirkka:
/gems/rollbar-2.8.2/lib/rollbar/rake.rb:23:in `patch?': undefined method `split' for nil:NilClass (NoMethodError)

Getting the above error with rake 11.0.1, looks like patch method is simply unable to determine what version of Rake is running.

Rollbar gem verison used is 2.8.2.

jondeandres:
Hi @pirkka, thanks for opening this issue. We hope to fix this this week.

Thanks.

loed-idzinga:
Running into the same issue, copied the output from the trace. 

```bash
** Invoke db:migrate (first_time)
** Invoke environment (first_time)
** Execute environment
rake aborted!
NoMethodError: undefined method `split' for nil:NilClass
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar/rake.rb:23:in `patch?'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar/rake.rb:6:in `patch!'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar/rake.rb:38:in `<top (required)>'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `require'
<path>s/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `block in require'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:240:in `load_dependency'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-4.2.5.2/lib/active_support/dependencies.rb:274:in `require'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar.rb:884:in `require_hooks'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar.rb:863:in `prepare'
<path>/2.3.0/lib/ruby/gems/2.3.0/gems/rollbar-2.8.2/lib/rollbar.rb:833:in `configure'
[...]

PR #397, opened 2016-03-09
Fix rake_version method

AmirMingaliev:
Hey, @jondeandres I've found the typo, which causes #396, I think. Could you kindly review this PR?

### DIFF
--- a/lib/rollbar/rake.rb
+++ b/lib/rollbar/rake.rb
@@ -28,8 +28,8 @@ module Rollbar
     def self.rake_version
       if Object.const_defined?('RAKEVERSION')
         return RAKEVERSION
-      elsif Rake.const_defined?('VERSION')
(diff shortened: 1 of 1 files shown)


jondeandres:
Thanks @AmirMingaliev, I'd say that was the cause of the problem yes :smile:. I'll merge and release a new version.

Thanks again!

stevenharman:
Phew, this just hit us after a seemingly mundane upgrade of a few dependencies. We're awaiting a new release of the `rollbar` gem. Thanks!

AmirMingaliev:
@jondeandres great!

lksv:
Should be nice to update also web pages which refer to last gem version:
* https://rollbar.com/docs/notifier/rollbar-gem/
* and the one which is showed when new project is created (at rollbar.com web )

PR #399, opened 2016-03-10
Namespace Rake out of Rollbar::Rake

abMatGit:
`Rake` is namespaced as `Rollbar::Rake` which causes `self.rake_version` to return `nil`. 

[As of 11.0.0](https://github.com/ruby/rake/commit/f593062cbd58a8d6a678d3b244bc6fdd2d4ef6f2) Rake removed `RAKEVERSION`.

### DIFF
--- a/lib/rollbar/rake.rb
+++ b/lib/rollbar/rake.rb
@@ -28,8 +28,8 @@ module Rollbar
     def self.rake_version
       if Object.const_defined?('RAKEVERSION')
         return RAKEVERSION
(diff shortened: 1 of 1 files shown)


jondeandres:
hi @abMatGit, we already have a PR for this, https://github.com/rollbar/rollbar-gem/pull/397. We'll merge it and release probably today.

abMatGit:
@jondeandres ah thanks for the notice, awesome!

PR #400, opened 2016-03-10
Fix rake development version to < 11

jondeandres:
We want to do this right now and remove this once we fix the development
dependencies problems we are having right now.


jondeandres:
This will make the tests to pass  in master when https://github.com/rollbar/rollbar-gem/pull/397 is merged

PR #401, opened 2016-03-12
Fix two doc bugs

paulrnash:
1) password_confirmation was listed twice
2) Example for ignoring an exception did not match text above it (said 404's, actually showed ignoring RecordNotFound)

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,6 @@ By default, the notifier will "scrub" the following fields from payloads before 
 - ```:password_confirmation```
 - ```:secret```
 - ```:confirm_password```
-- ```:password_confirmation```
 - ```:secret_token```
 
 And the following http header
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -434,7 +433,7 @@ If you'd like to customize this list, modify the example code in ```config/initi
 
 ```ruby
 config.exception_level_filters.merge!({
-  'ActiveRecord::RecordNotFound' => 'ignore',
+  'ActionController::RoutingError' => 'ignore',
```
jondeandres: Here you changed a not related line

paulrnash: No, please see the PR comment. This was the second doc bug I fixed. See how the doc text says "to ignore routing errors..." and then the example is for RecordNotFound instead.

jondeandres: Oh, sorry, you are right. I think the comment is confusing but right at same time, since if you raise a `ActiveRecord::RecordNotFound` exception in Rails you will have a 404 response code. I'll check this.

paulrnash: Ah, that may be true... but def not the only source of a 404. It kind of
depends on what you catch and/or coerce.

PR #405, opened 2016-03-21
Add threads reaper

amenzhinsky:
In some situations using threaded way of reporting errors when main process exits payloads are not delivered because all threads are killed straightaway.

For example rake tasks:
```ruby
Rollbar.use_threaded

task :foo do
  begin
    raise 'bar'
  rescue => error
    Rollbar.error(error)
  end
end
```

Another way is to add a method that would do something like this code, but it looks ugly for me: 
```ruby
begin
  tmp = Rollbar.configuration.use_async
  Rollbar.configuration.use_async = false
  yield
ensure
  Rollbar.configuration.use_async = tmp
end
```

### DIFF
--- a/lib/rollbar/delay/thread.rb
+++ b/lib/rollbar/delay/thread.rb
@@ -1,9 +1,64 @@
+require 'thread'
+require 'timeout'
+
 module Rollbar
   module Delay
     class Thread
-      def self.call(payload)
-        new.call(payload)
-      end
+      EXIT_SIGNAL  = :exit
+      EXIT_TIMEOUT = 3
+
+      Error        = Class.new(StandardError)
+      TimeoutError = Class.new(Error)
+
+      class << self
+        attr_reader :reaper
+
+        def call(payload)
+          spawn_threads_reaper
+          threads << new.call(payload)
+        end
+
+        private
+
+        def threads
+          @threads ||= Queue.new
+        end
+
+        def spawn_threads_reaper
+          return if @spawned
+          @spawned = true
+
+          @reaper ||= build_reaper_thread
+          configure_exit_handler
+        end
+
+        def build_reaper_thread
+          ::Thread.start do
+            loop do
+              thread = threads.pop
+
+              if thread == EXIT_SIGNAL
+                break
+              else
+                thread.join
+              end
+            end
+          end
+        end
+
+        def configure_exit_handler
+          at_exit do
+            begin
+              Timeout.timeout(EXIT_TIMEOUT) do
+                threads << EXIT_SIGNAL
+                reaper.join
+              end
+            rescue Timeout::Error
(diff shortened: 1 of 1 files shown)


amenzhinsky:
I'd like to discuss it and when it's ok I'll add some spec and do some refactoring if it's needed

jondeandres:
Hey @amenzhinsky, thank you very much for this PR. There is one point I'm concerned about: spawn the reaper always this class is loaded, even it's not used as async handler.

I think we should create this thread only if the `Rollbar::Delay::Thread` is being used. For this we can spawn the reaper thread if it's not spawned on first `.call()`. What do you think?

About refactoring, just take a look at https://codeclimate.com/github/rollbar/rollbar-gem/pull/405, since the spawn method is too long. We can split it into 3 methods: `spawn_threads_reaper`, `build_reaper_thread`, `configure_exit_handler` or similar.

We'll need to test this manually and we'll need specs, but this is a great work! Thanks

amenzhinsky:
Hi @jondeandres, I updated the PG, I followed all your advises except  `build_reaper_thread`, I added `::reaper` method that memoizes the reaper thread and call it in `spawn_threads_reaper` because I need access to it in `::configure_exit_handler` it's implicit a little. What do you think?

lib/rollbar/delay/thread.rb
```diff
@@ -1,9 +1,61 @@
...
+                  break
+                else
+                  thread.join
+                end
+              end
+            end
+        end
+
+        def spawn_threads_reaper
+          unless @spawned
```
jondeandres: According to the Ruby style, it's prefered not wrap a method inside a `unless` block. Here you can do `return if @spawned`

lib/rollbar/delay/thread.rb
```diff
@@ -1,9 +1,61 @@
...
+            @spawned = true
+
+            reaper # we spawn the reaper thread before configuring it
```
jondeandres: I'd prefer if here we have `@reaper||= build_reaper_thread`. And you use `attr_reader :reaper` in the class definition.

amenzhinsky:
@jondeandres done, please have a look

lib/rollbar/configuration.rb
```diff
@@ -44,6 +44,7 @@ class Configuration
     attr_accessor :uncaught_exception_level
     attr_accessor :scrub_headers
     attr_accessor :sidekiq_threshold
+    attr_accessor :thread_timeout
```
jondeandres: This name is confusing. Also, I'm not sure this should be something configurable. Maybe we can just use a constant for this.

jondeandres:
@amenzhinsky the PR is great now. I will need to play a little with this and check everything is working as expected.

Feel free to write and fix the specs for this.

Thanks!

Issue #411, opened 2016-03-23
undefined method `split' for nil:NilClass occur in assets:precompile


AknEp:
In my heroku enviroment, `undefined method `split' for nil:NilClass` was occurred in `assets:precompile`.

I found `Rake::VERSION` and `RAKEVERSION` is nil. so rollbar gem crashes in `Rollbar::Rake.patch!`

I don't know why Rake Version const was not defined ...
my Gemfile.lock has below.., and bundle install command success.

```
rake (11.1.1)
```

but, I think rollbar should be more robust.

jondeandres:
Hi @AknEp, this is resolved in lasts releases of the gem. Please try to update to the last one. THe fix is in this PR, https://github.com/rollbar/rollbar-gem/pull/397.

Rake 11.x was release the 9th o march and we fixed it on 10th. We know this may has caused problems to you, so we are sorry.

Thanks.

AknEp:
oops! Thank you for your quick response. sorry for my no-checking other issues.

but I think similar problem should occur, so #397 it not enough. I'll send PR.

PR #412, opened 2016-03-23
add nil check for rake.patch! for future robustness

AknEp:
refs #411

In the future, `Rake` class possibly change version API, this change prevent nil error in the future.

### DIFF
--- a/lib/rollbar/rake.rb
+++ b/lib/rollbar/rake.rb
@@ -20,6 +20,7 @@ module Rollbar
     end
 
     def self.patch?
+	  return false if rake_version.nil?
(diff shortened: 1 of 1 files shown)


AknEp:
I think `return true` is also okay, could someone judge this feature?

lib/rollbar/rake.rb
```diff
@@ -20,6 +20,7 @@ def self.skip_patch
     end
 
     def self.patch?
+	  return false if rake_version.nil?
```
jondeandres: Can you change this to `return false unless rake_version`.

Also, you have two many spaces in the indentation.

Thanks!

jondeandres:
This is merged by other PR, synced with master.

PR #414, opened 2016-03-24
Added support for CSP nonce when secure_headers.gem is used.

ajgon:
The `script_tag` method in Rollbar::Js::Middleware hardcoded the
<script> tag. This was causing problems, when `secure_headers.gem`
had CSP nonce enabled, and `'unsafe-inline'` disabled.

This patch fixes that, and also ensures that `Rollbar::Js::Middleware`
id loaded AFTER `SecureHeaders::Middleware` (to actually obtain the
nonce from the @env through `Rack::Request`).

### DIFF
--- a/lib/rollbar/js/frameworks/rails.rb
+++ b/lib/rollbar/js/frameworks/rails.rb
@@ -9,6 +9,24 @@ module Rollbar
         def prepare
           return if prepared?
 
+          if secure_headers?
+            insert_middleware_after_secure_headers
+          else
+            insert_middleware
+          end
+
+          self.prepared = true
+        end
+
+        def insert_middleware_after_secure_headers
+          instance = self
+
+          Rollbar::Railtie.initializer 'rollbar.middleware.js.frameworks.rails', :after => 'secure_headers.middleware' do |_app|
+            instance.insert_middleware
+          end
+        end
+
+        def insert_middleware
           require 'rollbar/js/middleware'
 
           config = {
@@ -16,8 +34,10 @@ module Rollbar
             :enabled => Rollbar.configuration.js_enabled
           }
           rails_config.middleware.use(::Rollbar::Js::Middleware, config)
+        end
 
-          self.prepared = true
+        def secure_headers?
+          defined?(::SecureHeaders)
         end
 
         def rails_config
--- a/lib/rollbar/js/middleware.rb
+++ b/lib/rollbar/js/middleware.rb
@@ -27,7 +27,7 @@ module Rollbar
       def _call(env, result)
         return result unless should_add_js?(env, result[0], result[1])
 
-        if response_string = add_js(result[2])
+        if response_string = add_js(env, result[2])
           env[JS_IS_INJECTED_KEY] = true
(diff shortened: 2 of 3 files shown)


lib/rollbar/js/frameworks/rails.rb
```diff
@@ -6,20 +6,38 @@ class Rails
...
-          return if prepared?
-
+        def self.insert_middleware
           require 'rollbar/js/middleware'
 
           config = {
             :options => Rollbar.configuration.js_options,
             :enabled => Rollbar.configuration.js_enabled
           }
-          rails_config.middleware.use(::Rollbar::Js::Middleware, config)
```
jondeandres: You don't need to change this line.

lib/rollbar/js/frameworks/rails.rb
```diff
@@ -6,20 +6,38 @@ class Rails
 
         alias prepared? prepared
 
-        def prepare
-          return if prepared?
-
+        def self.insert_middleware
```
jondeandres: Why this is a class method? It can be a instance method.

lib/rollbar/js/middleware.rb
```diff
@@ -117,6 +118,10 @@ def js_snippet
       end
 
       def script_tag(content)
+        if defined?(::SecureHeaders)
+          nonce = ::SecureHeaders.content_security_policy_script_nonce(@request)
+          return html_safe_if_needed("\n<script type=\"text/javascript\" nonce=\"#{nonce}\">#{content}</script>")
+        end
```
jondeandres: Can you add an extra line here?

lib/rollbar/js/middleware.rb
```diff
@@ -26,6 +26,7 @@ def call(env)
 
       def _call(env, result)
         return result unless should_add_js?(env, result[0], result[1])
+        @request = ::Rack::Request.new(env)
```
jondeandres: I'd prefer not adding more state to this class when `#call` is executed. You can see I've even passed the `env` to the different methods as argument. I'd like to maintain that style.

So just passing this object to `config_js_tag` and `snippet_js_tag` and then to `script_tag` will be enough.

lib/rollbar/js/middleware.rb
jondeandres: Could you change this to something like this?

```ruby
if defined?(::SecureHeaders)
  script_tag_content = ... #
else
  script_tag_content = ... #
end

html_safe_if_needed(script_tag_content)
```

jondeandres:
Hey @ajgon, thanks for this PR! It's awesome. I just left some comments about the style and minor issues. If you can change it'd be perfect.

Thanks!!

ajgon:
Hey @jondeandres, I'm glad you liked it :smile_cat: I fixed all the stuff you asked for, but if you have any additional remarks, just let me know!

Cheers!

jondeandres:
This is closed with https://github.com/rollbar/rollbar-gem/pull/437. Thanks @ajgon!

PR #421, opened 2016-04-08
Clean scope before every Sidekiq job execution

### LINKED ISSUE #416: Sidekiq 4.x scope side effects
I was doing a code review of this gem before adoption and I noticed that there isn't any code to clear thread locals between sidekiq jobs.  Since sidekiq 4.x (and maybe earlier?) reuse `Processor` threads to run more than one job (https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/processor.rb#L66-L68), if I use `Rollbar.scope!` to mutate the current scope I think there may be side effects for the next job.  I would expect to see sidekiq middleware that calls `Rollbar.reset_notifier!` before calling `yield`.

Note that the easiest way to test this would be to run sidekiq with only one `Processor` thread.

jondeandres:
For Sidekiq > 3.x we are using the error handlers for the error reporting while we use a middleware for Sidekiq < 3.x.

So, in order to clean the scope in both solutions, we add a middleware to do it for Sidekiq > 3.x and complete the existing middleware for Sidekiq < 3.x.

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -4,6 +4,14 @@ module Rollbar
   class Sidekiq
     PARAM_BLACKLIST = %w[backtrace error_backtrace error_message error_class]
 
+    class ClearScope
+      def call(worker, msg, queue)
+        Rollbar.reset_notifier!
+
+        yield
+      end
+    end
+
     def self.handle_exception(msg_or_context, e)
       return if skip_report?(msg_or_context, e)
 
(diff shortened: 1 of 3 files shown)


lib/rollbar/sidekiq.rb
```diff
@@ -28,12 +38,13 @@ def call(worker, msg, queue)
...
 
 Sidekiq.configure_server do |config|
-  if Sidekiq::VERSION < '3'
+  if Sidekiq::VERSION.split('.')[0].to_i < 3
     config.server_middleware do |chain|
       chain.add Rollbar::Sidekiq
     end
   else
-    config.error_handlers << Proc.new do |e, context|
+    chain.add Rollbar::Sidekiq::ClearScope
```
phlipper: `chain` is undefined here, see #422.

jondeandres: Thanks @phlipper, I'll release a new version in few minutes. Sorry for this.

PR #423, opened 2016-04-09
Fix Sidekiq support for version > 3.x

jondeandres:


paulmenzel:
Looking through the [change-log](https://github.com/rollbar/rollbar-gem/blob/master/CHANGELOG.md) it’d be great, if there would have been a commit message body, describing what the problem is, so that I know how big the problem is.

PR #424, opened 2016-04-09
Tiny patch to JS middleware injection

salamagd:
I noticed a small bug in `Rollbar::Js::Middleware#add_js` which, due to string slicing indices, causes the character immediately following the end of the `<head>` tag to be inserted once before the Rollbar JS script tags are inserted, as well as appearing immediately after the tags.

This is normally not noticeable/problematic if you've got some whitespace following the end of the `<head>` tag, like a carriage return, but if you don't (for example, using a minifier which strips out extraneous whitespace), you get this happening:

```
<head><
<script type="text/javascript">... config js tag ...</script>
<script type="text/javascript">... script js tag ...</script><meta ...>
```

(note the additional `<` that's been inserted after the head tag, which renders in the page)

Simple fix, as `head_open_end` contains the index of the character immediately following the `<head>`, we shouldn't include it in the first slice range.

### DIFF
--- a/lib/rollbar/js/middleware.rb
+++ b/lib/rollbar/js/middleware.rb
@@ -77,7 +77,7 @@ module Rollbar
         return nil unless head_open_end
 
         if head_open_end
-          body = body[0..head_open_end] <<
+          body = body[0..(head_open_end - 1)] <<
                  config_js_tag <<
                  snippet_js_tag <<
                  body[head_open_end..-1]


jondeandres:
Thanks for this @salamagd, I'll review the branch behavior and merge when done.

Thanks again!

jondeandres:
@salamagd I'm thinking maybe we should change this https://github.com/rollbar/rollbar-gem/pull/424/files#diff-60acb564885fb7ddb616f5f8e20b60a1R94 instead of apply your change

jondeandres:
Hey @salamagd, finally we merged https://github.com/rollbar/rollbar-gem/pull/435 completing the specs.

Thanks for this.

salamagd:
Many thanks, @jondeandres - appreciate it!

PR #425, opened 2016-04-11
Disable logging if Rollbar is disabled

### LINKED ISSUE #66: Feature request: disable logging in development if Rollbar is disabled
I have rollbar disabled in development, but any time there's an exception I get log messages like this:
```shell
[Rollbar] Reporting exception: undefined method `submissions' for nil:NilClass
[Rollbar] Exception not reported because Rollbar is disabled
```
It would be nice if there was a way to disable these as they clutter up my log file and are unnecessary if Rollbar is disabled.

jondeandres:
Close #66

### DIFF
--- a/lib/rollbar/logger_proxy.rb
+++ b/lib/rollbar/logger_proxy.rb
@@ -23,6 +23,8 @@ module Rollbar
     end
 
     def log(level, message)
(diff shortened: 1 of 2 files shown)


brianr:
LGTM

PR #430, opened 2016-04-13
Add delayed_job async handler

jondeandres:
I need to fix the tests for this, there are some issues requiring delayed_job here.

### DIFF
--- a/spec/rollbar/delay/delayed_job_spec.rb
+++ b/spec/rollbar/delay/delayed_job_spec.rb
@@ -1,0 +1,22 @@
+require 'spec_helper'
+
+require 'delayed_job'
+require 'delayed/worker'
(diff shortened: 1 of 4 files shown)


spec/rollbar/delay/delayed_job_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+require 'delayed_job'
+require 'delayed/worker'
+require 'rollbar/delay/delayed_job'
+
+describe Rollbar::Delay::DelayedJob do
+  before do
+h    Delayed::Worker.delay_jobs = false
```
brianr: typo? there's an extra `h` at the beginning of the line

PR #432, opened 2016-04-18
add sidekiq-specific framework and context params

heythisisnate:
This should make Rollbar more easily able to filter out errors coming from Sidekiq by setting the framework to "Sidekiq" and adding the `context` attribute to the error report similar to the way Bugsnag does by default.

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -16,7 +16,11 @@ module Rollbar
       return if skip_report?(msg_or_context, e)
 
       params = msg_or_context.reject{ |k| PARAM_BLACKLIST.include?(k) }
-      scope = { :request => { :params => params } }
+      scope = {
+        :request => { :params => params },
+        :framework => "Sidekiq: #{::Sidekiq::VERSION}"
+      }
+      scope[:context] = "sidekiq##{params['queue']}" if params.is_a?(Hash)
 
       Rollbar.scope(scope).error(e, :use_exception_level_filters => true)
     end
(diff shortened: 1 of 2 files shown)


brianr:
Very nice, thanks @heythisisnate !

This looks good to me. @jondeandres can you review as well?

heythisisnate:
Wow @brianr I didn't expect you to jump on this so quickly! I'm testing this out on a demo environment in my app first. Please wait until I confirm this is working the way I expect it to before merging.

brianr:
Sounds good.

One thing that won't work immediately is getting a sidekiq icon to appear in the Rollbar UI. That'll take a small change on our side, which we should be able to do quickly. Until then though you can still find these items by searching for `context:sidekiq` or `context:sidekiq#foo`.

heythisisnate:
I tested it briefly and this appears to do what I was expecting. The main
goal was to exclude Sidekiq errors when selecting Rails from the source
drop down on the /items page, and that appears to work. Yes, ideally I'd
like a Sidekiq option in that drop down, too. As well as being able to go
to /items?framework=sidekiq to show only Sidekiq errors.

If this gets merged, how quickly would you be able to publish a new gem and
add that?

jondeandres:
@heythisisnate these changes are nice, let me try this and we'll merge it. we hope to do it probably today.

Thanks!

heythisisnate:
Hey @jondeandres, thanks for merging this!
I was actually just about to reply with a little update but I saw this was merged already. After thinking about it a little more, I'm thinking that I'd prefer the `context` in Sidekiq errors to include the worker class instead of the queue. The goal here being so that it's easily searchable to find all errors that are occurring in a specific worker. Would you be open to that small change?

jondeandres:
@heythisisnate that has sense, probably the best option is set the job name to `context` and add `queue` also to the scope. we didn't release a new version yet, so feel free to open a PR for that.

heythisisnate:
I'll open a new PR by the end of the day.

brianr:
@heythisisnate Updated on the "icon on the Items page" part of this - that is done now, and will take effect for new items received starting ~5min ago.

PR #440, opened 2016-04-20
set the error context to the worker class name

heythisisnate:
@jondeandres @brianr as I mentioned in #432, on second thought I think putting the worker class name in the `context` parameter will be much more useful for searching & filtering. The Sidekiq queue name still comes through in `request.params.queue`, but this is less important for searching.

### DIFF
--- a/lib/rollbar/sidekiq.rb
+++ b/lib/rollbar/sidekiq.rb
@@ -20,7 +20,10 @@ module Rollbar
         :request => { :params => params },
         :framework => "Sidekiq: #{::Sidekiq::VERSION}"
       }
-      scope[:context] = "sidekiq##{params['queue']}" if params.is_a?(Hash)
+      if params.is_a?(Hash)
+        scope[:context] = params['class']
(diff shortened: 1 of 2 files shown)


lib/rollbar/sidekiq.rb
```diff
@@ -20,7 +20,7 @@ def self.handle_exception(msg_or_context, e)
         :request => { :params => params },
         :framework => "Sidekiq: #{::Sidekiq::VERSION}"
       }
-      scope[:context] = "sidekiq##{params['queue']}" if params.is_a?(Hash)
+      scope[:context] = params['class'] if params.is_a?(Hash)
```
jondeandres: Could you add the queue to `scope:queue]` here?

heythisisnate:
@jondeandres done!
Does the `queue` attribute do anything special? Is it searchable in the Rollbar UI?

jondeandres:
@heythisisnate not in the items page but in RQL section. but anyway I think it's interesting information to have.

Thanks :smile:

PR #443, opened 2016-04-22
Adds support for setting Net:HTTP open_timeout through config

bryandavisweb:
This allows you to configure a timeout for when your server can't talk to Rollbar and keeps your app from hanging.


jondeandres:
Hey @bryandavisweb, thanks for this. let me try this manually and maybe add some tests and we'll merge it.

PR #447, opened 2016-04-26
Added support for rollbar.js when SecureHeaders 2.x installed. Fixes #442

### LINKED ISSUE #442: Secure headers support doesn't compatible with secure_headers 2.x
Hi,

I updated rollbar to 2.10.0 and enable RollbarJS integration but `SCRIPT` tag will not inserted in response body. 😢 

The PR #437 sets `Rollbar::Railtie.initializer` after `secure_headers.middleware` but it's not exists in secure_headers 2.x.

* `railtie.rb` in secure_headers 3.0
https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/railtie.rb
* `railtie.rb` in secure_headers 2.x
https://github.com/twitter/secureheaders/blob/2.x/lib/secure_headers/railtie.rb

Although secure_headers 3.0 is a current major version, the authors says "2.x will be maintained" in README.

ajgon:
As a original author of the PR, I feel obliged to provide the fix.

However, `SecureHeaders 2.x` is nothing more than Rails controller hack, while Rollbar is a Rack Middleware. This means (and please correct me I'm wrong), the `rollbar.gem` doesn't have access to controller (or more exactly, controller instance variable `@content_security_policy_nonce` which holds the nonce) at the point when it invokes the `script_tag` method.

So what I did, I simply disabled the nonce fix for secure headers < 3 (or in more general, a non-middleware SecureHeaders version) - it's not perfect, but at least script tag is rendered properly again.

### DIFF
--- a/spec/rollbar/middleware/js_spec.rb
+++ b/spec/rollbar/middleware/js_spec.rb
@@ -97,7 +97,7 @@ END
         end
       end
 
-      context 'having a html 200 response and SecureHeaders defined' do
+      context 'having a html 200 response and SecureHeaders >= 3.0.0 defined' do
         let(:body) { [html] }
         let(:status) { 200 }
         let(:headers) do
@@ -106,6 +106,7 @@ END
 
         before do
           Object.const_set('SecureHeaders', Module.new)
+          SecureHeaders.const_set('VERSION', '3.0.0')
           allow(SecureHeaders).to receive(:content_security_policy_script_nonce) { 'lorem-ipsum-nonce' }
         end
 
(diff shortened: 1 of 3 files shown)


jondeandres:
Hey @ajgon, thank you for this. I'll try it.

We've added a plugins infrastructure to the gem and now some things have been moved. So now we have support for other libraries in `lib/rollbar/plugins/`. This is not released yet, but the new code should work with that.

Yesterday I found a problem with rollbar.js & secure headers integration, so I'll need to fix that. But I'll use your commit to add support for 2.0.

Thanks.

spec/rollbar/middleware/js_spec.rb
```diff
@@ -106,6 +106,7 @@
 
         before do
           Object.const_set('SecureHeaders', Module.new)
+          SecureHeaders.const_set('VERSION', '3.0.0')
```
jondeandres: These tests are ok, but the implementation doesn't check the version of SecureHeaders, no?

jondeandres: Ah ok, I see we had `allow(SecureHeaders).to receive(:content_security_policy_script_nonce) { 'lorem-ipsum-nonce' }`. Nice

PR #398, opened 2016-03-09
Untested adding comments to capistrano

Crisfole:
Somewhat WIP. It should work, but testing it proved to be more work than it was worth it if someone else can do it in a minute.

PR #460, opened 2016-05-03
Send rollbar_comment during deploys

wemcdonald:
Pretty self-explanatory.

FWIW I'm using this with the `capistrano-pending` gem to publish a changelog as the comment (I'm not on Github so I don't get the free commit log):

```
set :rollbar_comment, Proc.new { `cap #{fetch :stage} deploy:pending`.split("\n").grep(/^    /).join("\n\n") }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -847,6 +847,8 @@ set :rollbar_role, Proc.new { :app }
 Optionally, you can add a comment to your deploy with `rollbar_comment`. E.g.:
 
 ```ruby
+# Example: Interactively ask the user for a deploy comment.
+#   Alternativeely, you could generate a comment by (e.g.) querying your SCM repo
 set :rollbar_comment, Proc.new {
   ask :comment, "Describing this deploy"
   fetch(:comment)
--- a/README.md
+++ b/README.md
@@ -844,6 +844,15 @@ set :rollbar_env, Proc.new { fetch :stage }
 set :rollbar_role, Proc.new { :app }
 ```
 
+Optionally, you can add a comment to your deploy with `rollbar_comment`. E.g.:
+
+```ruby
+set :rollbar_comment, Proc.new {
+  ask :comment, "Describing this deploy"
+  fetch(:comment)
+}
+```
+
 NOTE: We've seen problems with Capistrano version `3.0.x` where the revision reported is incorrect. Version `3.1.0` and higher works correctly.
 
 ### Capistrano 2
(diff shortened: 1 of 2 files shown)


Crisfole:
Hey there, this is great. Thanks!

Can you add a test or two for this?

Once the tests pass we'll take a second look and hopefully merge it! It would fix my (poor attempt) in #398.

Crisfole:
Ok. So I'm told that adding Capistrano tests is challenging/difficult. What I'll do is test this manually once the other tests pass. :D

printercu:
Maybe it's better to make `current_revision` configured? I've redefined it with
```ruby
# deploy.rb
require 'rollbar/capistrano'

def current_revision
  `git log #{branch} -n 1 --pretty=format:"%H"`
end

```

jondeandres:
@wemcdonald I'm worried about this change cause it forces a user input, no? I guess some customers have automated deploys where they can't enter that value.

coryvirok:
I agree with Jon. Can we make this optional and not enabled by default?

jondeandres:
We can do this yes. I think we should first tests this manually since we don't have capistrano tests for now.

wemcdonald:
@jondeandres I just used the interactive case as an example since it seemed the most self-explanatory. See my first comment above for something non-interactive.

@coryvirok is your concern the same, or that we're including the key in the request with a null value? That doesn't seem like it's different from any of the other values...

coryvirok:
I was agreeing with Jon about not requiring users to enter a comment. After reading your README change more closely, it says that it's optional. Which is 👍 by me.

wemcdonald:
I just updated the text in the readme to describe what the example does and to provide another suggestion. I didn't include my use case because `deploy:pending` is no longer standard.

rokob:
I can't think of a natural default comment, but it is something that is easy to add if there is something obvious that I am missing. Otherwise, I think we should merge this.

mengqing:
+1, looking forward to get this feature integrated!! Also, just wondering if we can more control / configuration on what we can set as the comment?

Eg; we want to tag an environment variable on the remote server as a deploy comment once deployed

rivkahstandig3636:

ArturMoczulski:
LGTM. Merging.

@mengqing you should be able to easily configure the comment using
```
set :rollbar_comment, Proc.new {}
```

PR #461, opened 2016-05-04
Fix active_model require in validations plugin

supaspoida:
Loading the validations plugin while active_model is present without
active_record causes an exception.


jondeandres:
Hey @supaspoida, thanks for this, we missed to change the require when changing from active record to active model.

supaspoida:
@jondeandres big fan of rollbar, happy to help. and thanks for the snappy response!

dbackeus:
Another note here is that if `active_record` was present (within a rails app etc) but intentionally not loaded then it would be loaded her which could trigger things like `DatabaseCleaner` starting to use active record strategies instead of other intended orm strategies.

jondeandres:
This has been released in 2.11.2. Thanks again for the fix.

PR #467, opened 2016-05-12
add missing require

y-yagi:
Without this, `Rake::VERSION` can not be acquired, patch is always skip.


jondeandres:
Hey @y-yagi, thanks for this. I'll open a new PR to fix this since we need to rescue `LoadError` first.

y-yagi:
:+1: Good! I think that is a better approach.  I'll close this.

PR #473, opened 2016-05-24
Dont call as_json on Delayed::Backend::Sequel::Job

davidbegin:
instead call #to_hash, and turn all the keys to strings, to
  keep the rest of the implementation untouched

### DIFF
--- a/lib/rollbar/plugins/delayed_job/job_data.rb
+++ b/lib/rollbar/plugins/delayed_job/job_data.rb
@@ -6,7 +6,12 @@ class JobData
   end
 
   def to_hash
-    job_data = job.as_json
+    job_data = if job.respond_to?(:as_json)
+                 job.as_json
+               else
+                 Hash[job.to_hash.map { |k, v| [k.to_s, v] }]
+               end
+
     handler_parent = job_data['job'] ? job_data['job'] : job_data
     handler_parent['handler'] = handler_data
 


davidbegin:
We are seeing an error in Rollbar, when using Sequel and Delayed Job.
'undefined method `as_json'' for #<Delayed::Backend::Sequel::Job:0x007fe7f57d31f8>

In this file: https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/plugins/delayed_job/job_data.rb#L9

we are calling on @as_json on this class in Delayed Job Sequel
https://github.com/TalentBox/delayed_job_sequel/blob/56e7d69e1710b75ae7044a884961e1ebfd6a4661/lib/delayed/backend/sequel.rb

I updated Rollbar to just use #to_hash, if #as_json is not defined.

jondeandres:
Hey @davidbegin, thanks for this. Let me test it and we'll merge it.

lib/rollbar/plugins/delayed_job/job_data.rb
```diff
@@ -6,7 +6,12 @@ def initialize(job)
   end
 
   def to_hash
-    job_data = job.as_json
+    job_data = if job.respond_to?(:as_json)
```
jondeandres: I think we should extract this code into its own method. And also, we should ensure that `job` responds to `#to_hash`.

davidbegin: Thanks for reviewing this, I will work on those changes right now.

jondeandres:
Hey @davidbegin,

thank you very much for the notes. You've found a missing piece we forgot to move when we added the plugins architecture. So, `JobData` should be in the correct namespace and the `require` in the job_data spec should be fixed.

We'll do this change ourself and check that the specs are passing in master. After that, I think it's a good starting for your PR.

michaelirey:
@jondeandres Has there been any movement on this PR? Is there anything I could do to help?

Thanks!

PR #481, opened 2016-06-03
Update exception filter heading and TOC

agrberg:
Also cleaned up some self linking sections. GH will put a link icon over these when hovering and none of the other sections were self linked. Happy to pull out that commit if there was a reason for this.


jondeandres:
Thanks for this @agrberg!

PR #482, opened 2016-06-05
Remove MultiJson dependency.

printercu:
To use Oj add `Rollbar::JSON.oj!` in initializer.

There is no need in MultiJson while it's set up only to use Oj. If someone want to use any other adapter, he can just redefine `load & dump` methods in `Rollbar::JSON`.


printercu:
https://github.com/rails/rails/pull/10576 about multijson

jondeandres:
Hey @printercu, thank you very much for this. We'd consider removing multi_json dependency since it will not be supported any more. But I've found some issues and questions.

- Currently we support ruby 1.8, although we are thinking on stop the ruby 1.8 support soon. Which consequences will have this change on this?
- Now we detect the default adapter used by MultiJson and we always try to use Oj if it's found. With this change users would need to execute `Rollbar::JSON.oj!` in order to continue using Oj. This is a kind of interface breaking and probably it'd be better to run `oj!` ourself if `Oj` is found. Probably on the setup method we have.

What do you think?

Thanks.

printercu:
Second seems to be reasonable. I'll add this.
What about 1.8? Doesn't it have `json` in stdlib? I haven't used it for a long time.

printercu:
@jondeandres What setup method you've told about? Can't find any matching.

jondeandres:
oh, true @printercu. I think we removed it on a time when we didn't have multi_son. if you take a look at the history for lib/rollbar/json.rb you can find it. We used it to load oj when it was a depedency in the gem.

so, we'd need to setup on Rollbar initialization the library to be used.

printercu:
updated

jondeandres:
Hey @printercu, thanks. It looks pretty nice. I'll review it better, do some tests and figure out what is the scenario with ruby 1.8.

printercu:
@jondeandres Hi! any update on it?

jondeandres:
Hi @printercu, no yet sorry. We've been working on other stuff recently. We'll try to prioritize this in our next gem round :smile:.

Thank you very much for this, we really appreciate it.

printercu:
Cool, thank you!

rokob:
Ruby 1.8 does not come with the json gem bundled with the stdlib, that started with 1.9. The json gem supports 1.8, you just have to list it as a runtime dependency.

rokob:
1.8 also does not come with Rubygems

rokob:
This is a good change that we should merge, but we need to resolve the < 1.9 issues and remove the multi_json runtime dependency from the gemspec.

ArturMoczulski:
I'm closing this PR since there are conflicts with latest `master`. If you still want to changes from this PR, please sync with the latest master

PR #483, opened 2016-06-06
Add documentation for plugins architecture in docs/plugins.md

jondeandres:


docs/plugins.md
```diff
@@ -0,0 +1,46 @@
+# Plugins
+
+The plugins architecture offers an easy way to load the wrappers for libaries or frameworks we want to send Rollbar reports from. The different plugins we've written can be found in https://github.com/rollbar/rollbar-gem/tree/master/lib/rollbar/plugins. In this document we'll explain the interface that a plugin offers.
+
+In case you'd like to open a PR in this gem in order to add a plugins for a library or framework we still don't support, you should write a file wit the name of the framework you want to wrap in `lib/rollbar/plugins`. Here an example of a plugin:
+
+```ruby
+# lib/rollbar/plugins/my_framework.rb
+
+Rollbar.plugins.define('my-framework') do
```
brianr: Is it significant that this line uses a hyphen (`my-framework`) while the filename and the line below use an underscore (`my_framework`)?

jondeandres: The value passed to `define()` can be whatever string, `My Framework`, `MyFramework`, etc...I can change it if this is confused.

brianr: Gotcha. Is it used for anything?

jondeandres: Currently only on logging when the plugin fails to load, so the user can report exactly which plugin caused a problem.

brianr: OK, got it. It's kind of a nit-pick so I think we probably don't need to make any changes to these docs to describe it. (If I were just reading these docs, I would have followed the example code and everything would have been fine.)

PR #478, opened 2016-06-01
SecureHeaders fixes

jondeandres:

Issue #462, opened 2016-05-06
Configuration to disable Rollbar.js script tag nonce


spdawson:
Commit 984c88f4290bb914a24bed90a0e66d8afb5c6525 added a `nonce` to the Rollbar.js script tag, to work around an issue when using the `secure_headers` gem with CSP nonces enabled, and `'unsafe-inline'` disabled.

However, this has now created a problem in the case where nonces are disabled, and the CSP has `'unsafe-inline'` enabled. Many third party tools (e.g. New Relic's `newrelic_rpm` gem) inject nonce-free script tags into the page; to support these, it is necessary to add `'unsafe-inline'` to the CSP. Unfortunately, the presence of the nonce attribute on the Rollbar.js script tag causes the CSP to ignore the `'unsafe-inline'` setting, which means that any injected nonce-free script tags will not run.

I suggest the addition of a configuration flag, to enable the Rollbar.js nonce attribute to be disabled; presumably, it is a sensible default position for the nonce to be enabled.

jondeandres:
Hi @spdawson, thanks for open this issue. I'll take a look at this in next days. Thanks again and sorry for the inconveniences.

ajgon:
Hey @spdawson 

Maybe it's better idea to check that in SecureHeaders configuration itself? So if `'unsafe-inline'` is enabled, nonce will be disabled by default?

spdawson:
Thanks @ajgon; that looks like a nice solution which will "just work" from the perspective of an application that uses both Rollbar and `secure_headers`

PR #501, opened 2016-07-30
Configuration to disable script nonce

spdawson:
The auto-added script tag nonce attributes are not compatible with a secure_headers configuration in which `'unsafe-inline'` is present in the CSP for `script-src`. See #462.

Some work has been done in other PRs to address this issue --- see #478; this PR is a suggestion for a simplest-possible solution to the problem. Specifically, allow the Rollbar configuration to mandate that the nonce attribute should not be used.

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -116,14 +116,17 @@ module Rollbar
       end
 
       def script_tag(content, env)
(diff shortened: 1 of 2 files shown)


jondeandres:
@spdawson thanks for this, I think we should resume #478 and merge it

jondeandres:
This should be closed by https://github.com/rollbar/rollbar-gem/pull/478

PR #505, opened 2016-08-05
Send arguments with exception in active job plugin

PlugIN73:
Hi! We are using  active job plugin and we have troubles to understand what's wrong in our code without arguments of this job. 
In Active Job, [arguments](https://github.com/rails/rails/blob/master/activejob/lib/active_job/core.rb#L9) it's method like job_id that you are using

### DIFF
--- a/spec/rollbar/plugins/active_job_spec.rb
+++ b/spec/rollbar/plugins/active_job_spec.rb
@@ -9,6 +9,11 @@ describe Rollbar::ActiveJob do
     include Rollbar::ActiveJob
 
     attr_reader :job_id
+    attr_accessor :arguments
+
+    def initialize(*arguments)
+      @arguments = arguments
+    end
 
     def perform(exception, job_id)
       @job_id = job_id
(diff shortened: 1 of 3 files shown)


jondeandres:
@PlugIN73 Can you take a look at the tests? They are failing cause your change.

This looks good to me, thanks for the PR!

PlugIN73:
@jondeandres oh) I fixed tests. But this build was failed without my changes https://travis-ci.org/rollbar/rollbar-gem/jobs/150082517

jondeandres:
@PlugIN73 thanks for adding/fixing those tests! For the build is failing, we have a commit in another PR, https://github.com/rollbar/rollbar-gem/pull/503/commits/ad694bbb049734fbe30b7a28608c16e36e86d66f, which is still not merged. Would you mind do a `git fetch` and cherry-pick that commit? That should fix that build and your PR will be green :smile:.

Thanks!

PlugIN73:
@jondeandres done :)

PlugIN73:
@jondeandres ping :)

spec/rollbar/plugins/active_job_spec.rb
```diff
@@ -9,6 +9,11 @@ class TestJob
     include Rollbar::ActiveJob
 
     attr_reader :job_id
+    attr_accessor :arguments
+
+    def initialize(*arguments)
```
jondeandres: @PlugIN73 I guess with this you are giving the same behavior than for a real ActiveJob job. I have a doubt, I think `arguments` will return the arguments for the `perform` method, no?

Can you make this test more close to a real scenario?

rokob: https://github.com/rails/rails/blob/2a843b3538bbc88286b83562fe1bb517e01576f6/activejob/lib/active_job/core.rb#L56

rokob: This is how ActiveJob does it

rokob: Actually this is confusing, ActiveJob kinda does it this way, `peform_now` basically calls `perform(*arguments)` where arguments is set in init. So to get these to match, the arguments should match up with what is passed to perform because all of this is fake.

vtm9:
@PlugIN73 , maybe you resolve conflicts?

rivkahstandig3636:
@ArturMoczulski @jessewgibbs I'm just creating an issue for this and will link it here and close this PR.

ArturMoczulski:
This has now been merged in PR #772

PR #518, opened 2016-09-02
Send custom.orig_host and custom.orig_uuid on too large payloads

jondeandres:


lib/rollbar/item.rb
```diff
@@ -101,16 +101,24 @@ def dump
...
+      handle_too_large_payload(stringified_payload, result)
 
       nil
     end
 
+    def handle_too_large_payload(stringified_payload, final_payload)
+      original_size = Rollbar::JSON.dump(stringified_payload).bytesize
+      final_size = final_payload.bytesize
+      uuid = stringified_payload['data']['uuid']
+      host = stringified_payload['data']['server']['host']
```
brianr: If 'server' or 'host' are missing, will this cause an error? If so, we should wrap it in a begin-rescue. Here's how pyrollbar does it: https://github.com/rollbar/pyrollbar/blob/master/rollbar/__init__.py#L1252-L1259

jondeandres: Good point. That could happen on failsafe or internal error report, that don't add the server data, but having those reports to have large payload will be uncommon. But you are right, I'll fix this.

PR #519, opened 2016-09-06
Allow to override config

jondeandres:
Many users ask how to override the configuration for a specific block of
code. This PR adds that functionality through the method
`Rollbar.with_config`, that delegates to the per-thread
Rollbar::Notifier instance.

The argument passed to `#with_config` is a `Hash` having key names equal
to the allowed configuration options and the new value to use for the
given block of code.

Example:

```ruby
Rollar.configure do |config|
  config.environment = 'production'
end

Rollbar.with_config(environment: 'staging') do
  begin
    # do something that will crash
  rescue => e
    Rollbar.error(e)
  end
end
```

In the example from above, the reported error will use `staging` as
environment instead of `production`.


spec/rollbar/notifier_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+    let(:new_config) do
+      { 'environment' => 'foo' }
+    end
+
+    it 'creates a new notifier with merged scope and configuration' do
+      new_notifier = subject.scope(new_scope, new_config)
+
+      expect(new_notifier).not_to be(subject)
+      expect(new_notifier.configuration.environment).to be_eql('foo')
```
coryvirok: Can you also make sure that the parent notifier's configuration has not been changed?

jondeandres: Nice, done.

brianr:
Looks great 👍

PR #521, opened 2016-09-07
Increase max payload from 128kb to 512kb

jondeandres:


lib/rollbar/truncation.rb
```diff
@@ -9,7 +9,7 @@ module Rollbar
   module Truncation
     extend ::Rollbar::Truncation::Mixin
 
-    MAX_PAYLOAD_SIZE = 128 * 1024 # 128kb
+    MAX_PAYLOAD_SIZE = 1024**2 # 1MB
```
coryvirok: Let's not go this high. How about 512 KB?

jondeandres: Ok, cool. I'll change it

PR #522, opened 2016-09-09
Send GET, POST and raw body in their correct place

jondeandres:
We are sending the GET, POST, JSON body and the Rails request parameters in `params`. According to our API specification we can send `GET`, `POST`, `body` and `params` keys in the request data.

This PR just fix that sending the correct data in the correct places.


brianr:
lgtm 👍

PR #535, opened 2016-09-29
Fix wrong version number in Readme

florrain:
Hi team!

There must have been a wrong copy/pasta during the last release routine. This fixed the last version number in the changelog.


jondeandres:
awesome, thanks @florrain! I think we should definitely write a script to create the release :smile:

PR #547, opened 2016-11-23
Add note about requiring hashrocket syntax

danielwestendorf:
I spent some time figuring out why my errors weren't being sent off to Rollbar. It turns out defining the queue as `config.use_sidekiq queue: 'default'` isn't valid. Hashrocket syntax is required.

### DIFF
--- a/lib/generators/rollbar/templates/initializer.rb
+++ b/lib/generators/rollbar/templates/initializer.rb
@@ -55,6 +55,7 @@ Rollbar.configure do |config|
   # Enable delayed reporting (using Sidekiq)
   # config.use_sidekiq
   # You can supply custom Sidekiq options:
+  # Hashrocket syntax required
   # config.use_sidekiq 'queue' => 'default'
 
   # If you run your staging application instance in production environment then


brianr:
Is the real issue that `'queue'` needs to be a string, not a symbol?

danielwestendorf:
Yes, I believe that's the case. I'm not sure if that's an expectation of Sidekiq or a requirement of this gem. Either way, a little comment might be helpful to others since many rubyists may naively try to use rails/ruby 1.9+'s keys-as-symbols syntax.

jondeandres:
When using `sidekiq_options` method to configure the job class, you can use `queue: 'value'` or `:queue => 'value`. But in this case we use `Sidekiq::Client.push` to push the job to the queue. That method needs a `String` key for `queue`. Documentation from the Sidekiq code:

```
    # The main method used to push a job to Redis.  Accepts a number of options:
    #
    #   queue - the named queue to use, default 'default'
    #   class - the worker class to call, required
    #   args - an array of simple arguments to the perform method, must be JSON-serializable
    #   retry - whether to retry this job if it fails, true or false, default true
    #   backtrace - whether to save any error backtrace, default false
    #
    # All options must be strings, not symbols.  NB: because we are serializing to JSON, all
    # symbols in 'args' will be converted to strings.
    #
    # Returns nil if not pushed to Redis or a unique Job ID if pushed.
    #
    # Example:
    #   push('queue' => 'my_queue', 'class' => MyWorker, 'args' => ['foo', 1, :bat => 'bar'])
    #
```

@danielwestendorf in this case, I think the comment you added is not 100% correct. For example:

```ruby
'queue' => 'value' # this is hashrocket

:queue => 'value' # this is also hashrocket
```

So the problem is not that the interface needs to use hashrocket syntax, but passing a `String` key for `queue`.

danielwestendorf:
Okay, yes, this PR is incorrect.

Perhaps the existence of this PR will be enough to lead confused users to a resolution.

jondeandres:
Thanks anyway @danielwestendorf!!

PR #549, opened 2016-12-08
Handle new exception message structure from sidekiq >= 4.2.3

jkubacki:
## Problem
`config.sidekiq_threshold = x` doesn't work for sidekiq >= 4.2.3
In 4.2.3 sidekiq changed message format passed to `#handle_exception`. Now it always have format of `{context: 'Context', job: {...}}`.

## Solution
When sidekiq version is >= 4.2.3, fetch :job from message.

Made it work also with older sidekiq versions and wrote specs for this method.


jondeandres:
Hey @jkubacki, thanks a lot for this. We'll need to test it manually and we will need to have the tests passing, but it's a great work so far.

Thanks!

gtd:
Looks like it just needs rebasing to pick up the build fix from https://github.com/rollbar/rollbar-gem/pull/554...

jkubacki:
There is better solution for this issue here #598.

PR #550, opened 2016-12-14
Removed doctoc tag

jessewgibbs:
It's causing issues in our GH pages submodule setup.  Switching the page to use the same format as other notifiers.


brianr:
The downside of this change is that it means the TOC will no longer be visible when viewing the readme inside GitHub. Are we ok with that?

As an alternative, maybe we could make _all_ of the notifiers have the generated TOC in them (like this library did before this PR), and make whatever changes are needed so that it appears OK on rollbar.com/docs ?

PR #556, opened 2016-12-19
Fix validations plugin on Rails 5.0 with belong_to

### LINKED ISSUE #534: Problem with rails validations @ 2.13.2
The rails 5 default validation for `belongs_to` being required(unless you use `optional: true`) stop working when updating from `2.12.0` to `2.13.2`

``` ruby
class PhoneNumber < ApplicationRecord
  belongs_to :phone_book
end
```

this test fails using `2.13.2` and works using `2.12.0`

``` ruby
require 'rails_helper'

RSpec.describe PhoneNumber, type: :model do
  describe "validation" do
    it "requires a phone_book" do
      pn = PhoneNumber.new
      expect(pn.valid?).to be false
      expect(pn.errors).to have_key(:phone_book)
    end
  end
end
```

jondeandres:
fixes #534

### DIFF
--- a/lib/rollbar/plugins/validations.rb
+++ b/lib/rollbar/plugins/validations.rb
@@ -30,13 +30,13 @@ Rollbar.plugins.define('active_model') do
       include Rollbar::ActiveRecordExtension
     end
 
-    # If ActiveRecord::Base has been already loaded,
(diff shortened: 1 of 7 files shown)


plentz:
👏

PR #558, opened 2016-12-20
Add Rollbar::Middleware::Rack

jondeandres:
Convert `Rollbar::Middleware::Rack` to be a class instead of a module,
so it can be added to the Rack middleware chain with `use
Rollbar::Middleware::Rack`.

Also, move the code from `Rollbar::Middleware::Sinatra` to
`Rollbar::Middleware::Rack` since Sinatra behavior is a subset of Rack
behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,28 +62,13 @@ $ heroku config:add ROLLBAR_ACCESS_TOKEN=POST_SERVER_ITEM_ACCESS_TOKEN
 That's all you need to use Rollbar with Rails.
 
 
-### Rack
+### Sinatra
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -77,28 +77,22 @@ end
...
   # ...
 end
 ```
 
-Then mount the middleware in your app, like:
+If you disabled the `Rack::Builder` monkey patch or it doesn't work for the Rack framework you are using, then add our Rack middleware to your app:
 
 ```ruby
-require 'rollbar/middleware/sinatra'
+require 'rollbar/middleware/rack
```
rivkahstandig3636: missing an end '

PR #560, opened 2017-01-02
Add host configuration option

### LINKED ISSUE #540: custom hostname
We run AWS EC2 instances and their hostnames aren't very descriptive, so we have a custom hostname variable that we use for syslog logging, identification, etc.

Currently, when rollbar sends a message it uses `Socket.gethostname` to get the hostname, but I'd like to be able to set a custom hostname, when configuring rollbar: i.e.

```
Rollbar.configure do |config|
  config.hostname = 'an-alternate-hostname'
end
```

Looks like it's easy enough to change the `server_data` method to check the config for the hostname variable?  Happy to discuss alternative implementations too.

jondeandres:
Allows using a custom value for the host name. Usage:

```ruby
Rollbar.configure do |config|
  config.host = 'my-custom-host'
end
```
Close #540

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -131,6 +131,12 @@ tracked by [rollbar-agent](https://github.com/rollbar/rollbar-agent).
 Indicates which framework you're using. Common options include 'Rails',
 'Sinatra', and 'Rack' to name a few.
 
+### host
+
+**Default** `nil`
+
+The hostname (reported to Rollbar as `server.host`). When nil, the value of `Socket.gethostname` will be used.
+
 ### ignored_person_ids
 
 **Default** `[]`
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -131,6 +131,12 @@ tracked by [rollbar-agent](https://github.com/rollbar/rollbar-agent).
 Indicates which framework you're using. Common options include 'Rails',
 'Sinatra', and 'Rack' to name a few.
 
+### host
+
+**Default** unspecified
+
+Allows using a custom host value instead the returned by `Socket.gethostname`.
```
brianr: How about:

```
### host

**Default** nil

The hostname (reported to Rollbar as `server.host`). When nil, the value of `Socket.gethostname` will be used.
```

jondeandres: :+1:

jondeandres: Changed

brianr:
👍

PR #567, opened 2017-02-09
Ensure params filtering take precedence over hash/array objects

carlosantoniodasilva:
This matches the current Rails implementation of parameter filtering, where it is possible to set a filter that will hide an entire array or nested hash, instead of just the inner keys.

Consider the following params payload:

```ruby
    user: { username: "foo", password: "bar" }
```

Currently we can set a filter for both "username" and "password", but setting for the entire "user" params hash does not work, it's simply ignored, because we parse hashes separately looking at each individual key first - before checking the configured filter.

With this change, setting a filter for "user" would effectively hide it from the payload, scrubbing the field as expected, since we first check the filter before proceeding with other processing.

---

Unrelated: I'm getting this failure locally on both master and this branch:

```
  1) Rollbar project_gems should handle regex gem patterns
     Failure/Error: gem_paths.length.should > 1

       expected: > 1
            got:   1
     # ./spec/rollbar_spec.rb:1173:in `block (3 levels) in <top (required)>'
```

I haven't stopped to check it properly, but doesn't seem to be related to the changes here since it's also happening on master.

Let me know if I can help with anything!

### DIFF
--- a/lib/rollbar/scrubbers/params.rb
+++ b/lib/rollbar/scrubbers/params.rb
@@ -53,13 +53,15 @@ module Rollbar
         return scrub_array(params, options) if params.is_a?(Array)
 
         params.to_hash.inject({}) do |result, (key, value)|
-          if value.is_a?(Hash)
+          if fields_regex && fields_regex =~ Rollbar::Encoding.encode(key).to_s
+            result[key] = Rollbar::Scrubbers.scrub_value(value)
+          elsif value.is_a?(Hash)
             result[key] = scrub(value, options)
           elsif value.is_a?(Array)
             result[key] = scrub_array(value, options)
           elsif skip_value?(value)
             result[key] = "Skipped value of class '#{value.class.name}'"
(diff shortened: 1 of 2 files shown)


jondeandres:
Hi @carlosantoniodasilva, thanks for the PR. It's quite good at all!

Tests are failing in Travis because a problem between our gemfiles and changes in rubygems. We'll fix master and merge a new PR with your commits, I think that will be the easiest.

carlosantoniodasilva:
@jondeandres sounds good, thanks! If you want I can also merge master back here when it's resolved there, just let me know.

rokob:
I just pushed a fix for the tests to master. Could you merge master back into your branch?

carlosantoniodasilva:
@rokob thanks, I've rebased from master and CI is running again.

rokob:
This looks good to me, thanks!

tt:
@rokob, can you release this?

rokob:
Yeah I will try to get a release out today.

PR #570, opened 2017-02-24
Scrub keys from hash arguments

aabes:
This is an attempt to filter out large elements from a body of a POST message, that rollbar is attempting to post in the content of an error.

The environment is a rack app (running in puma). A post request with a structure similar to below causes an error - 

```
 {
   "foo": 1,
   "blob": "very large data"
 }
 
```

The rollbar configuration has:
 
config.scrub_fields = [:blob]

but despite that, blob is included in the payload (and posting the error fails due to size).

This PR ensures that the hash keys included in the scrub_fields are excluded from the payload.

Does this ring right?

### DIFF
--- a/lib/rollbar/scrubbers/params.rb
+++ b/lib/rollbar/scrubbers/params.rb
@@ -52,14 +52,18 @@ module Rollbar
 
         return scrub_array(params, options) if params.is_a?(Array)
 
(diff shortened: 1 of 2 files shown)


rokob:
I am going to close this because https://github.com/rollbar/rollbar-gem/pull/567 which was just merged solves this problem.

aabes:
@rokob - what version of the gem will this equivalent PR be available in (i.e. I want to switch away from my fork back to a supported version).

PR #571, opened 2017-02-24
try to fix rbx gemfile issues

rokob:
Bundler doesn't do what I would have expected for the platform flag. It still requires the Gemfile to be resolvable even if that is not possible on a particular platform. The problem with rubysl and rbx is that rubysl requires ruby ~>2.0 which doesn't work when running tests with a <2.0 ruby version.


rokob:
This is just a restructuring of Gemfiles to get tests to pass, nothing should be functionally different, so merging.

Issue #563, opened 2017-02-02
queue name when async through dj


inquisitivus:
Hi!
Is there any way to specify delayed_job queue name for async submissions? 
Something like `config.use_resque :queue => 'my_queue'` for Resque.

jondeandres:
Hey @inquisitivus, thanks for writing. For resque you can just use `config.use_resque` for the default queue, or you can set the queue as you suggest: `config.use_resque :queue => :my_queue`.

jondeandres:
@inquisitivus sorry, I though you were asking for resque and not delayed_job.

Currently we don't support named queues for DJ, we'll do this soon. Until we fix this and release a new version of the gem, you can have the same behavior using this snippet:

```ruby
require 'rollbar/delay/delayed_job'

ROLLBAR_DJ_QUEUE = 'rollbar'

Rollbar::Delay::DelayedJob.instance_eval do
  def call(payload)
    new.delay(:queue => ROLLBAR_DJ_QUEUE).call(payload)
  end
end
```

Ping us if you have any problem with it and we'll help you. Thanks!

inquisitivus:
Thanks!

PR #574, opened 2017-03-06
specify delayed_job queue

inquisitivus:
Hi!
This is for my question in #563.

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -145,8 +145,11 @@ module Rollbar
       self
     end
 
-    def use_delayed_job
+    def use_delayed_job(options = {})
       require 'rollbar/delay/delayed_job'
+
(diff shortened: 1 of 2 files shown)


rokob:
This will certainly break the existing behaviour where people assume jobs are run on the default delayed_job queue. It should be possible to add this feature, but doing it this way won't work.

rokob:
One approach would be to add an option similar to https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/configuration.rb#L163 in the configuration for delayed_job here https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/configuration.rb#L148, and then if a queue is set on the delayed job class you can use that in the call method otherwise use the current default way of doing it.

inquisitivus:
I've changed my bits according to your advice. Is it ok now?

rokob:
Looks good to me, thanks!

PR #579, opened 2017-03-24
Replace present? call with plain ruby alternative

### LINKED ISSUE #578: Reporter relies on ActiveSupport but it's not a dependency
This code calls `present?` which is from ActiveSupport (https://github.com/rollbar/rollbar-gem/blob/368445ac9c15f883cd92eb427458ff1173ae1c51/lib/rollbar/item/frame.rb#L108):

```ruby
file_lines[from_line, (to_line - from_line + 1)].select(&:present?)
```

We fixed this in our project by bundling ActiveSupport. Should rollbar-gem include this as a dependency?

jondeandres:
Fixes #578

### DIFF
--- a/lib/rollbar/item/frame.rb
+++ b/lib/rollbar/item/frame.rb
@@ -105,7 +105,8 @@ module Rollbar
         to_line = lineno - 2
         from_line = [to_line - MAX_CONTEXT_LENGTH + 1, 0].max
 
-        file_lines[from_line, (to_line - from_line + 1)].select(&:present?)
+
+        file_lines[from_line, (to_line - from_line + 1)].select { |line| line && !line.empty? }
       end
     end
   end
--- a/lib/rollbar/item/frame.rb
+++ b/lib/rollbar/item/frame.rb
@@ -105,7 +105,6 @@ module Rollbar
         to_line = lineno - 2
         from_line = [to_line - MAX_CONTEXT_LENGTH + 1, 0].max
 
-
         file_lines[from_line, (to_line - from_line + 1)].select { |line| line && !line.empty? }
       end
     end


lib/rollbar/item/frame.rb
```diff
@@ -105,7 +105,8 @@ def pre_data(file_lines, lineno)
         to_line = lineno - 2
         from_line = [to_line - MAX_CONTEXT_LENGTH + 1, 0].max
 
-        file_lines[from_line, (to_line - from_line + 1)].select(&:present?)
+
+        file_lines[from_line, (to_line - from_line + 1)].select { |line| line && !line.empty? }
```
rokob: So the two cases are `line == nil` or `line` a string. If `line` is nil then the first half of the conditional fails and we do not select it, otherwise `line` will have an `empty?` method, so `!line.empty?` will be equivalent to `present?`. Active support implements `present?` as `!blank?` and `blank?` as `responds_to(:empty?) ? !!empty? : !self` which is maybe a little bit safer because it works on any object not just something with an `empty?` method defined. But we can assume here that `line` is of a primitive type like string, array, hash that has an `empty?` method defined.

rokob:
LGTM

lib/rollbar/item/frame.rb
jondeandres: yah, in this method `file_lines` is an array of strings. I think I did this to not send empty strings on the `pre` section. But fixing this I've seen we are not doing it for the `post`. For now I think this just fixes the issue, but this can be maybe improved.

v-yarotsky:
@jondeandres thanks for the fix! Would it be possible to merge this any time soon? I'd like to switch one of our projects to Rollbar, but it doesn't use activesupport, and I'd rather keep it that way :)

PR #582, opened 2017-04-01
If multiple forwarded hosts are present in the headers, use the first

ismith:
This addresses a case where "HTTP_X_FORWARDED_HOST" = "api.projectalloy.org, api.projectalloy.org", resulting in an error scrubbing the url.

(Hosting provider is NearlyFreeSpeech.net.)


rokob:
LGTM

PR #585, opened 2017-04-13
Set revision at rollback

threetreeslight:
When Capistrano does not set the current revision at the time of
rollback, an item that there is no revision is notified to the rollbar.

### DIFF
--- a/lib/rollbar/tasks/rollbar.cap
+++ b/lib/rollbar/tasks/rollbar.cap
@@ -31,6 +31,15 @@ namespace :rollbar do
 end
 
 namespace :deploy do
+  task :get_current_revision do
+    on primary fetch(:rollbar_role) do
+      within release_path do
+        set(:current_revision, capture(:cat, 'REVISIONS'))
+      end
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


lib/rollbar/tasks/rollbar.cap
```diff
@@ -31,6 +31,15 @@ namespace :rollbar do
 end
 
 namespace :deploy do
+  task :get_current_revision do
+    on primary fetch(:rollbar_role) do
+      within release_path do
+        set(:current_revision, capture(:cat, 'REVISIONS'))
```
rokob: Should probably also redirect STDERR to /dev/null so that current_revision doesn't get set if the REVISIONS file is missing for some reason.  Also, isn't the file named 'REVISION' and not 'REVISIONS'?

```ruby
set(:current_revision, capture(:cat, 'REVISION', '2>/dev/null'))
```

My knowledge of Capistrano is a bit lacking, does this file always exist or is it only for Capistrano 3+?

ArturMoczulski:
Closing the PR for now due. Reopen when there will be feedback @threetreeslight

PR #588, opened 2017-04-19
Add Configuration#use_exception_level_filters option

### LINKED ISSUE #587: allow configuring exception level filters:true/false as a global setting for manual reporting
Hi there, 
Currently if we manually report an issue through Rollbar.error(e), if we want Rollbar to take into consideration our exception level filters, we need to add a silly flag (:use_exception_level_filters => true).

Either cancel this flag, or allow for us to set it globally as a rollbar configuration setting. 

This issue just creates dirty code on our end..

jondeandres:
Let the users to set a global `use_exception_level_filters` option

Usage:

```
Rollbar.configure do |config|
  config.use_exception_level_filters = true
end
```

By default the setting is `false` so we keep the current interface

Fix #587

### DIFF
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -565,6 +565,16 @@ describe Rollbar do
 
           Rollbar.error(exception)
         end
+
+        it 'should not use the filters if overriden at log site' do
+          Rollbar.configure do |config|
+            config.exception_level_filters = { 'NameError' => 'ignore' }
+          end
+
+          Rollbar.error(exception, :use_exception_level_filters => false)
+
+          expect(Rollbar.last_report[:level]).to be_eql('error')
+        end
       end
     end
 
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -536,6 +536,82 @@ describe Rollbar do
       Rollbar.error(exception).should == 'disabled'
     end
 
+    context 'using configuration.use_exception_level_filters_default' do
+      before do
+        Rollbar.configure do |config|
+          config.use_exception_level_filters_default = true
+        end
+      end
+
+      context 'without use_exception_level_filters argument' do
+        it 'sends the correct filtered level' do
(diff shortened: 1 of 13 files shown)


rokob:
This seems fine to me. I guess the only question is whether you want this global configuration setting to be overridable on a per-call basis, i.e. if `use_exception_level_filters` is true in the configuration, based on this code, you cannot add `:use_exception_level_filters => false` on a particular error call and expect that to work. I think that is okay and conforms to what I would expect by setting the global flag. So I'd say go  ahead and merge this, once you fix the tests.

brianr:
Hm I think I would expect that:

- the new flag @jondeandres added is the default
- the default can be overridden by passing a param in the call

What if we do something like:

- rename the new config param to `use_exception_level_filters_default`. Its default will be `false` to maintain the current behavior.
- if you don't pass the `:use_exception_level_filters` param in a call, then the value of the config param is used. (maybe do this by having the param default to `nil`)
- if you do pass the `:use_exception_level_filters` param, then the value you provide is used

jondeandres:
@brianr @rokob I've pushed force here with Brian's suggestions.

spec/rollbar_spec.rb
```diff
@@ -536,6 +536,82 @@
...
+        it 'sends the correct filtered level' do
+          Rollbar.configure do |config|
+            config.exception_level_filters = { 'NameError' => 'warning' }
+          end
+
+          Rollbar.error(exception)
+
+          expect(Rollbar.last_report[:level]).to be_eql('warning')
+        end
+
```
rokob: This is missing a test when `config.use_exception_level_filters_default = true` has been set, and when `:use_exception_level_filters => false` is passed in with the `Rollbar.error` call.

rokob:
I squashed those commits.

spec/rollbar_spec.rb
rokob: I added one

Issue #472, opened 2016-05-24
Exception from Delayed::Job not reported if job fails to deserialize


adamniedzielski:
If Delayed::Job fails to deserialize a job it throws `DeserializationError`. This happens early in the process and the hook `lifecycle.around(:invoke_job, &Delayed::invoke_job_callback)` is not called.

Looking at the code it is rescued here https://github.com/collectiveidea/delayed_job/blob/1e679174b513dc711f7d764c8bd9d4f5c6495276/lib/delayed/worker.rb#L235-L237 and the only hook it calls is "failure".

Do you have any idea how we can account correctly for this case?

jondeandres:
@adamniedzielski do you know the reasons of the `DeserializationError`? Do you think that has a relation with the gem itself? Or is just that you'd like to report this errors?

adamniedzielski:
@jondeandres Thanks for your response!

The ```DeserializationError``` comes from the Delayed Job and has no relation with the Rollbar gem. I would like this ezception to be reported by Rollbar, because it prevents the job from executing and this is something I definitely want to be aware of :smile:

adamniedzielski:
Quick way to reproduce the problem:
```ruby
"abc".delay.first(->{})
Delayed::Worker.new.run(Delayed::Job.last)
```

I think that now I understand why this gem uses ```around(:invoke_job)``` rather than ```after(:error)``` or ```after(:failure)``` - the exception is not available for these callbacks.

In order to have a basic reporting for jobs which failed even before the first execution we would have to add something similar to:

```ruby
class RollbarPlugin < ::Delayed::Plugin
  callbacks do |lifecycle|
    lifecycle.around(:invoke_job, &Delayed::invoke_job_callback)
    lifecycle.after(:failure) do |job|
      ::Rollbar.error("Delayed::Job ID=#{job.id} failed to execute") if job.attempts.zero?
    end
  end
end
```

jondeandres:
Hey @adamniedzielski, thanks for this, we'll try to release a new version with this soon.

adamniedzielski:
Thanks @richvorp! 💚 💯 🎉

yoonwaiyan:
Really hope to get this released asap, delayed job failure is high in a server under my monitor.

PR #592, opened 2017-05-30
Fixes: https://github.com/rollbar/rollbar-gem/issues/472

richvorp:
If there is every an error within the payload object itself, within delayed jobs, rollbar does not propagate that error, since delayed job fails to call the :invoke_job lifecycle hook and instead calls the :failure lifecycle hook.

### DIFF
--- a/spec/rollbar/plugins/delayed_job_spec.rb
+++ b/spec/rollbar/plugins/delayed_job_spec.rb
@@ -35,10 +35,14 @@ describe Rollbar::Delayed, :reconfigure_notifier => true do
   end
 
   context 'with failed deserialization' do
+    let(:expected_args) do
+      [/Delayed::DeserializationError/, {:use_exception_level_filters=>true}]
+    end
+
     it 'sends the exception' do
       expect(Rollbar).to receive(:scope).with(kind_of(Hash)).and_call_original
       allow_any_instance_of(Delayed::Backend::Base).to receive(:payload_object).and_raise(Delayed::DeserializationError)
-      expect_any_instance_of(Rollbar::Notifier).to receive(:error)
+      expect_any_instance_of(Rollbar::Notifier).to receive(:error).with(*expected_args)
 
       FailingJob.new.delay.do_job_please!(:foo, :bar)
     end
--- a/spec/rollbar/plugins/delayed_job_spec.rb
+++ b/spec/rollbar/plugins/delayed_job_spec.rb
@@ -34,6 +34,15 @@ describe Rollbar::Delayed, :reconfigure_notifier => true do
     end
   end
 
+  context 'with failed deserialization' do
+    it 'sends the exception' do
+      expect(Rollbar).to receive(:scope).with(kind_of(Hash)).and_call_original
+      allow_any_instance_of(Delayed::Backend::Base).to receive(:payload_object).and_raise(Delayed::DeserializationError)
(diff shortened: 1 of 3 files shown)


coryvirok:
Thanks @richvorp! @rokob can we get this merged and released?

richvorp:
There is still a potential for the error to not fire if https://github.com/richvorp/rollbar-gem/blob/delayed_job_failure_hook/lib/rollbar/plugins/delayed_job/plugin.rb#L48 check fails and the dj_threshold is higher than 0.

An alternative would be to make it always error with:
```
data = Rollbar::Delayed.build_job_data(job)
::Rollbar.scope(:request => data).error(job.last_error, :use_exception_level_filters => true) if job.last_error
```
instead of https://github.com/richvorp/rollbar-gem/blob/delayed_job_failure_hook/lib/rollbar/plugins/delayed_job/plugin.rb#L14

I can push that change if you guys agree.

jondeandres:
@richvorp I think it would be ok to skip the report on that case if the number of attempts is less than the threshold, that would be the expected behavior, I guess.

spec/rollbar/plugins/delayed_job_spec.rb
```diff
@@ -34,6 +34,19 @@ def do_job_please!(a, b)
...
   end
 
+  context 'with failed deserialization' do
+    let(:expected_args) do
+      [/Delayed::DeserializationError/, {:use_exception_level_filters=>true}]
+    end
+
+    it 'sends the exception' do
+      expect(Rollbar).to receive(:scope).with(kind_of(Hash)).and_call_original
+      allow_any_instance_of(Delayed::Backend::Base).to receive(:payload_object).and_raise(Delayed::DeserializationError)
```
jondeandres: I guess this call to `#payload_object` is not the one we do in `JobData#handler_data` no? I say it because you are rescuing that one.

richvorp: It is the same call. The problem happens when the payload object itself cannot be deserialized (https://github.com/collectiveidea/delayed_job/blob/master/lib/delayed/backend/base.rb#L74) and it was also called in handler_data to construct the job details.

spec/rollbar/plugins/delayed_job_spec.rb
```diff
@@ -34,6 +34,19 @@ def do_job_please!(a, b)
...
+      expect_any_instance_of(Rollbar::Notifier).to receive(:error).with(*expected_args)
+
+      FailingJob.new.delay.do_job_please!(:foo, :bar)
```
rokob: If I follow this test correctly, this line will cause delayed job to eventually call `do_job_please!`, the backend will call `payload_object` during the call to `JobData.new(job).to_hash`, this is stubbed to raise a deserialization error, which is rescued and therefore `to_hash` is able to return something with `handler_parent['handler']` therefore set to `{}`. The way things are set up with the plugin means that delayed job will make a failure lifecycle call, which we hook into and call report which is where we satisfy the expectation of the notifier receiving a call to `error` with the deserialization error as argument. A failure to deserialize the payload causes delayed job to short circuit to failing the job and never retrying so we have to report it this way.

rokob:
The problem still remains that we will not report this issue if the threshold is greater than 0 because delayed_job will never retry if there is a deserialization error.

rokob:
Let's merge this, and then make a separate issue for handling the threshold > 0 case.

PR #598, opened 2017-06-09
support new sidekiq context structure

itszootime:
Hope I'm not causing any trouble by basically duplicating https://github.com/rollbar/rollbar-gem/pull/549. However, I recognised that this pull request has gone stale. There's also a little bit more here:

* A simpler implementation, not requiring the construction of two `Gem::Version` instances per error handled.
* The specs now test with arguments the handler can expect to receive from Sidekiq.
* An (extremely?) rare edge case where [Sidekiq calls handle_exception with a `nil` context](https://github.com/mperham/sidekiq/blob/v3.5.4/lib/sidekiq/launcher.rb#L34) is now handled correctly.

It'd be great if we can get this merged as the `sidekiq_threshold` behaviour is currently completely broken when using Sidekiq >= 4.2.3. Cheers!

### DIFF
--- a/lib/rollbar/plugins/sidekiq/plugin.rb
+++ b/lib/rollbar/plugins/sidekiq/plugin.rb
@@ -12,10 +12,11 @@ module Rollbar
       end
     end
 
-    def self.handle_exception(msg_or_context, e)
-      return if skip_report?(msg_or_context, e)
+    def self.handle_exception(ctx_hash, e)
+      job_hash = ctx_hash[:job] || ctx_hash
+      return if skip_report?(job_hash, e)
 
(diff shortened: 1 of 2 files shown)


coryvirok:
Thanks @itszootime! We will take a look soon.

itszootime:
I've noticed that this flags a lot of quality issues on Codacy, am happy to implement the suggested improvements if necessary.

jkubacki:
@coryvirok could you take a look at it guys? We don't want another solution to go stale ;)

rokob:
I'll take a look today

rokob:
This looks good to me, merging it in

PR #602, opened 2017-06-13
Support person data in rollbar.js

and-megan:


and-megan:
@jondeandres This works but it's automatic, ie you don't need to list it in the config.js_options, - should this be optional or is it ok as is?

and-megan:
I made it optional, but I can take it out

lib/rollbar/middleware/js.rb
```diff
@@ -103,7 +107,12 @@ def close_old_response(response)
       end
 
       def config_js_tag(env)
-        script_tag("var _rollbarConfig = #{config[:options].to_json};", env)
+        js_config = config[:options].dup
+        if js_config[:payload][:person_tracking]
```
rokob: If `:payload` does not exist in the config then this throws

lib/rollbar/middleware/js.rb
```diff
@@ -103,7 +107,12 @@ def close_old_response(response)
...
+          person_data = extract_person_data_from_controller(env)
+          js_config[:payload][:person] = person_data if person_data
```
rokob: Same here

and-megan:
@rokob I think it's fixed now

lib/rollbar/middleware/js.rb
```diff
@@ -103,7 +107,12 @@ def close_old_response(response)
       end
 
       def config_js_tag(env)
-        script_tag("var _rollbarConfig = #{config[:options].to_json};", env)
+        js_config = config[:options].dup
+        if js_config.try(:[], :payload).try(:[], :person_tracking)
```
jondeandres: I think we just need to do:

```ruby
person_data = extract_person_data_from_controller(env)
if person_data
  js_config[:payload] ||= {} # We initialize the `payload` dictionary just with `{}` if it doesn't exist
  js_config[:payload][:person] = person_data
end
```

and-megan:
@jondeandres Looks good. I had :person_tracking so that someone could turn rollbar.js person tracking on or off, rather than have it be automatic.

jondeandres:
Mmmm @and-megan, we even don't have that option in the gem itself, so I think it will be ok sending that information automatically. Using an external option would be the way to do it.

PR #605, opened 2017-06-16
Add 2.3.4 and 2.4.1 rubies to .travis.yml

timcraft:


jondeandres:
Hi @timcraft, thanks for opening this PR. I think there are tests failing because the mix of ruby version and rails version is not compatible. We would need to skip those in the matrix.exclude section in travis.yml

timcraft:
@jondeandres Makes sense in principle but looks like a lot of the jobs are failing before the test run. For example the ruby 2.4.1 and rails30.gemfile job here:

https://travis-ci.org/rollbar/rollbar-gem/jobs/243617130

That's failing because of an error installing oj 2.12.14--is there any reason why oj is pinned to that older version? (Latest released version is 3.1.3)

rivkahstandig3636:

ArturMoczulski:
Thanks @timcraft for the PR, but I'm closing it due to conflicts with the latest in `.travis.yml`. If you still want to see these changes, please submit another PR

PR #626, opened 2017-08-10
Adds proxy server support.

petelacey:


petelacey:
Bump.  Any feedback on this PR?

lib/rollbar/notifier.rb
```diff
@@ -496,6 +500,40 @@ def do_post(uri, body, access_token)
...
+      px.password = proxy[:password]
+      px
+    end
+
+    def proxy_from_env(uri)
+      uri.find_proxy
+    end
+
+    def null_proxy
+      OpenStruct.new(:host => nil, :port => nil, :user => nil, :password => nil)
```
rokob: Why OpenStruct and not just a plain struct? i.e.

`Struct.new(:host, :port, :user, :password).new`

petelacey: To be honest, this started life as a Struct, then an abandoned refactoring changed it to an OpenStruct and I neglected to change it back.  Happy to do that.  It makes this code more concise.

rokob:
This seems like a reasonable change to me.

lib/rollbar/notifier.rb
```diff
@@ -8,6 +8,7 @@
 require 'rollbar/delay/thread'
 require 'rollbar/logger_proxy'
 require 'rollbar/item'
+require 'ostruct'
```
printercu: This one is redundant.

PR #627, opened 2017-08-16
Use :local_user variable

abicky:
Capistrano 3.3.3 introduced `:local_user` and I think `:rollbar_user` should refer to this value.
In addition, this change allow users to change `:rollbar_user` by changing `:local_user`.


rokob:
This seems like a good change to me, it is backwards compatible because it is the same behaviour if `:local_user` is not defined.

PR #634, opened 2017-09-05
Add the option to stop sending non-project frames

hazefully:
I ran into a problem in my current company while integrating Rollbar with JIRA; the backtrace frames are sent too large which make the automatically generated bugs on JIRA unreadable because all the backtrace frames are sent, whether they are non-project or in-project frames. I thought it would be a nice feature if the sending of non-project frames can be controlled by the user.

User can decide to either send non-project exception frames, which is the default, or not to send them so it can be easier to pinpoint the code that is causing the exception precisely. 

This feature is configurable using config.send_extra_frame_data which can be `true` or `false` based on the user's preference. Example:

```ruby
Rollbar.configure do |config|
  config.send_non_project_frames = false
end

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,6 +643,15 @@ Rollbar.silenced {
   foo = bar  # will not be reported
 }
 ```
+## Non-project frames
+
+By default all backtrace frames are sent whether they are in-project frames or non-project frames. Since the backtrace can be very long and non-project frames are mostly irrelevant, you can configure Rollbar to only send the in-project frames by setting `send_non_project_frames` to either `true` or `false`. Example:
+
+```ruby
+Rollbar.configure do |config|
+   config.send_non_project_frames = true
+end
+```
 
 ## Sending backtrace without rescued exceptions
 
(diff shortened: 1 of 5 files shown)


rokob:
So the implementation looks reasonable and I can understand the need for this. The only alternative to this would be to do this manually via a user set transform. This does seem to be a widely used enough use case to include in the main codebase. So my opinion is to merge this. Anyone else have an opinion?

brianr:
Thinking about the root issue (that the backtrace is hard to read in a JIRA ticket), I think a better solution would be to fix this on the Rollbar side (in the product), rather than in the gem.

We already have a way to know which frames are project vs. non-project, so we'd just need a setting (or maybe changing the default behavior) so that the stacktrace in JIRA tickets and such only shows in-project frames.

@hazefully would that solve your problem?

hazefully:
@brianr I believe this will indeed fix the problem of unreadability, but it will need to be to implemented for all integrations similar to JIRA. (I have no idea how this works in the Rollbar product itself)

PR #641, opened 2017-09-28
Add option to only send report on failure

bmalinconico:
I had a desire to silence exceptions from my jobs that inevitably self-corrected.

This PR adds a configuration option to only send the report to Rollbar when the job gets marked as failed.

Another possible enhancement would be to add an optional callback to the job which accept the exception and make a skip decision itself. This pattern is used in DJ so each one can have its own max_attempts, or custom back-off function: https://github.com/collectiveidea/delayed_job/#custom-jobs

### DIFF
--- a/lib/rollbar/plugins/delayed_job/plugin.rb
+++ b/lib/rollbar/plugins/delayed_job/plugin.rb
@@ -53,7 +53,10 @@ module Rollbar
     end
 
     def self.skip_report?(job)
-      job.attempts < ::Rollbar.configuration.dj_threshold
+      job.attempts < ::Rollbar.configuration.dj_threshold && (
+        !::Rollbar.configuration.dj_only_report_failures ||
+          !job.failed_at
+      )
     end
 
     def self.build_job_data(job)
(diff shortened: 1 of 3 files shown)


lib/rollbar/plugins/delayed_job/plugin.rb
```diff
@@ -53,7 +53,10 @@ def self.report(e, job)
     end
 
     def self.skip_report?(job)
-      job.attempts < ::Rollbar.configuration.dj_threshold
+      job.attempts < ::Rollbar.configuration.dj_threshold && (
```
rokob: The name `only_report_failures` doesn't make sense to me. The only case that is different is when the number of attempts is less than the threshold, only_report_failures is true and the job has failed. Before this change that would have been a skip, and with this change that would be a report. So we are actually reporting failures under the threshold, but we are still reporting any time the number of attempts is greater or equal to the threshold regardless of whether the job failed or not.

ArturMoczulski:
I'm closing this PR since @rokob comment is valid and there has been no response from @bmalinconico 

I'll report when there are further comments.

PR #647, opened 2017-11-02
Insert the snippet after meta charset/content-type tags if either is present

### LINKED ISSUE #546: Middleware inserts JavaScript before charset meta tag
According to e.g. [this](https://www.w3.org/International/questions/qa-html-encoding-declarations.en), the `charset`-declaring `<meta>` tag must occur entirely within the first 1024 bytes of the document.

However, the Rollbar middleware inserts the JavaScript tags immediately after the opening `<head>` tag; this causes validation errors for HTML documents that include the auto-inserted Rollbar JavaScript tags.

rokob:
Fixes #546

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -92,9 +92,9 @@ module Rollbar
       end
 
       def find_insertion_point(body)
-        find_end_after_regex(body, /<meta\s*charset=/) ||
-        find_end_after_regex(body, /<meta\s*http-equiv="[Cc]ontent-[Tt]ype"/) ||
-        find_end_after_regex(body, /<head\W/)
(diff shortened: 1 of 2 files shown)


lib/rollbar/middleware/js.rb
```diff
@@ -91,9 +91,15 @@ def build_body_with_js(env, body, head_open_end)
           body[head_open_end + 1..-1]
       end
 
-      def find_end_of_head_open(body)
-        head_open = body.index(/<head\W/)
-        body.index('>', head_open) if head_open
+      def find_insertion_point(body)
+        find_end_after_regex(body, /<meta\s*charset=/) ||
```
brianr: Any reason not to make all three of these regexes fully case-insensitive?

PR #651, opened 2017-11-09
add a whitelist option for the fields to scrub

### LINKED ISSUE #614: scrub_fields too broad
when added a param to scrub fields such as "name", every field that had the substring called "name" was obfuscated. so something like "first_name" or "last_name" etc would be obfuscated. Can this be more specific?

rokob:
Fixes #614 

Still needs an update to the README but code is working

We already do exact string matching for url parameters but not for params. Not sure why that is, but it means we don't need a whitelist for url parameters technically.

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -41,6 +41,7 @@ module Rollbar
     attr_accessor :js_enabled
     attr_accessor :safely
     attr_accessor :scrub_fields
+    attr_accessor :scrub_fields_whitelist
     attr_accessor :scrub_user
     attr_accessor :scrub_password
     attr_accessor :user_ip_obfuscator_secret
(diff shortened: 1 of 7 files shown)


rokob:
No the point is to supplement the existing options because the elements of `scrub_fields` get turned into regular expressions so that something like `name` matches `username`. The behaviour you describe comes if you use `scrub_all` and this whitelist.

jondeandres:
aha, great. thanks for explaining @rokob!

AndrewRayCode:
I'm having trouble understanding if it's possible to whitelist fields in Rollbar with this Gem, for example I'd like to remove `session.*` but also whitelist `sesssion.useful_key`. It's not documented that this gem supports Rails's `filter_parameters` config, and it's unclear if this gem supports the (also undocumented) behavior or `filter_paramaters` that can take a lambda and do something with the result as one of the array entries (https://github.com/rails/rails/blob/ce4d467f7c2fc812e257a87bd4875c1f1f08a981/railties/test/application/configuration_test.rb#L335-L337)

rivkahstandig3636:
@ArturMoczulski

ArturMoczulski:
`scrub_whitelist` has been already implemented and is available since `v2.16.4`. Closing the PR.

PR #665, opened 2018-01-17
Propose more resilient boot crash logging

bf4:


rokob:
This seems like an okay change to me, anyone else have an opinion? It is just a doc change, but it makes sense.

bf4:
@rokob Thanks for the quick review.  This is the code we're using now. Our app didn't boot last night and we didn't get any notifications.  This is one of our fixes to address that :)

bf4:
@rivkahstandig3636 🌈

bf4:
I just had a sidekiq boot failure which rollbar didn't successfully notify me of.. due to sidekiq failing, and came back to the repo to look for where it might be like this solution, but see it was removed in https://github.com/rollbar/rollbar-gem/pull/752 ...

Issue #601, opened 2017-06-13
Issue with Capistrano3 deploys and proxy settings


and-megan:
We likely need code in one of the following files:

https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/capistrano.rb
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/capistrano3.rb

so that the proxy settings can be pulled from environment vars and used in the Net request.

tomsparks:
Looks like there are two ways for this to be handled - either from the rollbar config in the deploy.rb for the project or checking to see if there's an environment variable set. Given the variety of issues I've had with capistrano actually taking in environment variables, its probably better to use the ones set in deploy.rb

tomsparks:
I dug into this a bit more - Net::HTTP will actually detect if the "http_proxy" environment variable is set and automatically use it. This may be a capistrano issue instead, I'm looking into it but the environment variable is set on my side.

tomsparks:
Yes, this is still an issue.

PR #666, opened 2018-01-18
Passing proxy vars manually

htmldrum:
Current cap tasks doesn't support env vars. Passes nil into Net::HTTP.start (https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#method-c-start) which, after inspecting the source , will sends nils in for proxy variables.

### DIFF
--- a/lib/rollbar/tasks/rollbar.cap
+++ b/lib/rollbar/tasks/rollbar.cap
@@ -21,7 +21,12 @@ namespace :rollbar do
       request      = Net::HTTP::Post.new(uri.request_uri)
       request.body = ::JSON.dump(params)
 
-      Net::HTTP.start(uri.host, uri.port, :use_ssl => true) do |http|
+      p_env = ENV['http_proxy'] || ENV['HTTP_PROXY']
+      p_uri = p_env ? URI.parse(p_env) : nil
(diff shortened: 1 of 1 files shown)


htmldrum:
#601

amasses:
👏 This was driving me nuts, but clearly I didn't put the effort in to dig into the cause.

lib/rollbar/tasks/rollbar.cap
```diff
@@ -21,7 +21,12 @@ namespace :rollbar do
       request      = Net::HTTP::Post.new(uri.request_uri)
       request.body = ::JSON.dump(params)
 
-      Net::HTTP.start(uri.host, uri.port, :use_ssl => true) do |http|
+      p_env = ENV['http_proxy'] || ENV['HTTP_PROXY']
+      p_uri = p_env ? URI.parse(p_env) : nil
```
rokob: This should be in a `begin/rescue` block to catch the `URI::InvalidURIError` that is possible to be thrown here. Furthermore, `p_env` truthy does not imply `p_uri` is not nil below, so you need to guard against that in the calls to `host` and `port`.

rivkahstandig3636:

Issue #664, opened 2018-01-16
Middleware only renders Rollbar snippet on pages with 200 status code


NoahFisher:
### Context: 
We use statically generated error pages in rails to render custom errors. We use Rollbar to report custom errors on those pages (504 timeouts are the most useful error we log with this setup as we only have to generate the error page once, and serve the static version thereafter). 

### Problem: 
The rollbar middleware only injects the snippet on pages with a 200 html response, which means we have to copy/paste the snippet (and all subsequent gem updates). 

Suspect code in this file (line 9 contains a comment stating this is expected behavior and line 43 contains the suspect code). 
`.../ruby-2.2.2/gems/rollbar-2.15.5/lib/rollbar/middleware/js.rb`
```
def add_js?(env, status, headers)
  enabled? && status == 200 && !env[JS_IS_INJECTED_KEY] &&
  html?(headers) && !attachment?(headers) && !streaming?(env)
end
```

Any ideas for a workaround? Is there a stable way to render the snippet manually in specific layouts?

rivkahstandig3636:
Thanks for the report! We'll investigate.

NoahFisher:
I used a similar approach outlined in the comments of #569.

ArturMoczulski:
@NoahFisher I agree, the snippet should be rendered on any HTML response regardless of the status code. Here is a PR with your suggested change: https://github.com/rollbar/rollbar-gem/pull/670

ArturMoczulski:
This has been fixed in PR https://github.com/rollbar/rollbar-gem/pull/684

PR #670, opened 2018-01-24
GitHub Issue #664: Middleware only renders Rollbar snippet on pages with 200 status code

ArturMoczulski:
Rollbar.js snippet should be rendered in any HTML response bodies, regardless of the status code of the response. Previously it was only rendered on 200 response codes.

This PR addresses https://github.com/rollbar/rollbar-gem/issues/664


ArturMoczulski:
This PR was cleaned up in PR 684

PR #675, opened 2018-02-01
Bug undefined blank for string in Rollbar::Logger

### LINKED ISSUE #674: Rollbar::Logger uses active_support/core_ext without depending on it
Rollbar does not have ActiveSupport as
a runtime dependency in gemspec and it is
not required in Rollbar::Logger but is used.

way to reproduce:

```irb
require 'rollbar/logger'
Rollbar::Logger.new.error("foo")
  # NoMethodError: undefined method `blank?' for "foo":String
```

can be worked around by requiring
`active_support/core_ext/object/blank`

```irb
require 'rollbar/logger'
require 'active_support/core_ext/object/blank'
Rollbar::Logger.new.error("foo")
  # => "disabled"
```

duksis:
Possible solution to close #674 - feel free to use it or not to :)

Implements a local blank for usage without core_ext
Check if message responds to :blank?, if it does not
use a local implementation of it.

local implementation taken from
https://github.com/rails/rails/blob/v5.1.4/activesupport/lib/active_support/core_ext/object/blank.rb#L17

### DIFF
--- a/lib/rollbar/logger.rb
+++ b/lib/rollbar/logger.rb
@@ -28,7 +28,7 @@ module Rollbar
 
       message ||= block_given? ? yield : progname
 
-      return true if message.blank?
+      return true if blank?(message)
 
       rollbar.log(rollbar_level(severity), message)
     end
@@ -64,6 +64,11 @@ module Rollbar
     end
 
     private
+
+    def blank?(message)
+      return message.blank? if message.respond_to?(:blank?)
+      message.respond_to?(:empty?) ? !!message.empty? : !message
+    end
 
     # Find correct Rollbar level to use using the indexes in Logger::Severity
     # DEBUG = 0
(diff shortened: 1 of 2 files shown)


ArturMoczulski:
@duksis why not just add a dependency on `active_support/core_ext` in the `Gemfile` rather than duplicating their implementation?

duksis:
@ArturMoczulski it's always good to keep your dependencies as minimal as possible.
At the moment `rollbar-gem` has only one dependency in gemspec and in order to reuse `active_support/core_ext` instead of copying this one line we would be needed to add the whole `activesupport` as a runtime dependency.

at the end it's a decision that the maintainers have to make: 

4 lines with a guard clause and 1 line copied from core ext 
**VS** 
371 KB of `activesupport` code in dependencies

ArturMoczulski:
@duksis that's a good point

Issue #656, opened 2017-11-30
Security vulnerability with Rails dependency


coryvirok:
![screen shot 2017-11-30 at 2 59 04 pm](https://user-images.githubusercontent.com/769323/33459539-2ddb2e52-d5df-11e7-96fa-fad72c90b120.png)

brianr:
Isn't this just used for the tests?

coryvirok:
It's used in various parts of the actual SDK, if I'm reading this correctly - https://github.com/rollbar/rollbar-gem/tree/master/lib/rollbar/plugins/rails

brianr:
That just monkeypatches Rails if it exists, I believe. @jondeandres may be able to clarify.

I think the `Gemfile` that is committed here might not even be used... it's generated by Appraisal, which actually generates the whole set of gemfiles for various combinations of dependencies which we test on TravisCI.

ArturMoczulski:
@brianr the Gemfile that's committed to current `rollbar/rollbar-gem:master` uses `rails/rails:4.2.3`.

Here are more details on vulnerabilities in this version:
https://www.cvedetails.com/vulnerability-list/vendor_id-12043/product_id-22568/version_id-190893/Rubyonrails-Ruby-On-Rails-4.2.3.html

I'm not sure what we should do about this. Upgrade to `rails/rails:4.2.7.1`?

ArturMoczulski:
This is resolved in https://github.com/rollbar/rollbar-gem/pull/688

PR #688, opened 2018-02-22
GitHub Issue #656: replace rails 4.2.3 with 4.2.7.1

ArturMoczulski:


ArturMoczulski:
On `ruby-1.9.3` in some builds `URL::Generic.find_proxy` wasn't getting loaded. This resulted in a range of failures and eventually an infinite loop in async tests causing the build to time out. I forced rollbar load `open-uri` which invludes `URL::Generic.find_proxy` in older Ruby versions.

PR #702, opened 2018-03-21
Follow symlinks when looking up source maps

alexdunae:
The capistrano task wasn't following symlinks to public/assets and public/packs directories -- the `find -L` flag allows it to do so


ArturMoczulski:
Thanks @alexdunae

PR #715, opened 2018-04-19
Don't send person's personally identifiable information as a default

johnsyweb:
#### Context

Sending a user's email address to a third-party is not desirable under The EU General Data Protection Regulation ([GDPR](https://www.eugdpr.org/)) and sending it to Rollbar is not a sensible default.

#### Change

The default is *not* to send the email address but how to do so remains a documented option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,7 +379,7 @@ If the methods to extract the ```id```, ```username```, and ```email``` from the
 ```ruby
 Rollbar.configure do |config|
   config.person_id_method = "user_id"  # default is "id"
-  config.person_username_method = "user_name"  # default is "username"
+  config.person_username_method = "user_name"  # default is `nil`
   config.person_email_method = "email_address"  # default is `nil`
 end
 ```
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ If the methods to extract the ```id```, ```username```, and ```email``` from the
 Rollbar.configure do |config|
   config.person_id_method = "user_id"  # default is "id"
   config.person_username_method = "user_name"  # default is "username"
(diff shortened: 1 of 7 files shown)


johnsyweb:
I am told that Rollbar are [preparing for the EU General Data Protection Regulation, and will support these new regulations by the May 2018 deadline](https://rollbar.com/compliance/). This is great.

My PR is more fundamental than this, in that application users have not given **explicit** [consent](https://www.eugdpr.org/the-regulation.html) to transmit this personally identifiable information (PII) to Rollbar in the first place. 

Indeed, developers don't explictly grant the Rollbar gem permission to transmit email addresses to Rollbar. Talking with developers both inside and outside of the EU I could find nobody who would use an email address value for debugging purposes -- with the exception of when it was the processing of the email address which caused the error. In any case, the system calling the Rollbar gem would have this information and there is no need to transmit it.

denmat:
I fully support your initiative @johnsyweb, I would also support going further and making `username` and other identifiable information opt in.

rokob:
We are discussing this internally as we would like to ensure some consistency for how we handle this across all of our notifier libraries. We will update this PR once we have come to some kind of consensus. I understand where you are coming from, but I'd rather not get into a legal debate on this PR at this time.

To be clear, for anyone who comes across this, at the moment one can easily opt-out of the collection of said data through the configuration setting for the person methods. This discussion is only about whether the default should be opt-in or opt-out.

johnsyweb:
Thanks @rokob!

> This discussion is only about whether the default should be opt-in or opt-out.

💯

ArturMoczulski:
This can be merged now.

Issue #695, opened 2018-03-04
SecureHeaders deprecation


brian-kephart:
Ruby 2.5.0, Rails 5.2.0rc1

```
~/.rvm/gems/ruby-2.5.0/bundler/gems/rollbar-gem-8613b275368c/lib/rollbar/middleware/js.rb:160:in `append_nonce?': [DEPRECATION] `#get` is deprecated. It will be removed in the next major release. Use SecureHeaders::Configuration.dup to retrieve the default config.
```

Deprecated method appears here:
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/js.rb#L160

jondeandres:
This can be closed, https://github.com/rollbar/rollbar-gem/pull/730.

PR #729, opened 2018-04-26
update `append_nounce?` method for SecureHeaders

marunbai:
This pr try to solve this issue: https://github.com/rollbar/rollbar-gem/issues/695

According to http://www.rubydoc.info/gems/secure_headers/SecureHeaders/Configuration#default-class_method, the `get` method is returning default config without any arg so we could just use `default` instead of `get`


marunbai:
I moved this pr into a company repo

Issue #726, opened 2018-04-26
Only collect person.id by default for person tracking


rivkahstandig3636:
Allow the ability to send person.email and person.username, but by default we should only collect person.id going forward.

@ArturMoczulski This is a top priority.

ArturMoczulski:
This is addressed by https://github.com/rollbar/rollbar-gem/pull/715

PR #734, opened 2018-05-04
GitHub Issue #726: Only collect person.id by default for person tracking

ArturMoczulski:
PR addressing https://github.com/rollbar/rollbar-gem/issues/726

By default, we don't send any Person Data to Rollbar. However, this can be turned on with config option `person_collect_email_username` which will start collecting the current user's email and username.

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -31,6 +31,7 @@ module Rollbar
     attr_accessor :person_id_method
     attr_accessor :person_username_method
     attr_accessor :person_email_method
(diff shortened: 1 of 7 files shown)


rokob:
I'm in favor of the style of doing this as in #715 where people can use the existing configuration options to pick which data to submit. This approach gives you an all or nothing on email/username and also introduces new configuration options when we already have the username/email method options. I think we can just merge that one.

ArturMoczulski:
Closing this in favor of https://github.com/rollbar/rollbar-gem/pull/715

PR #739, opened 2018-05-17
[Docs] Mention #default_logger is Rails.logger when using Rails project

f3ndot:
Per:

https://github.com/rollbar/rollbar-gem/blob/137aed9014772404dfb0c8a6dab18adc3983ded2/lib/rollbar/plugins/rails/railtie_mixin.rb#L13-L21

`Rollbar.default_logger` will return Rails' configured logger instance instead of `Logger.new(STDERR)`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -59,7 +59,7 @@ end
 
 ### default_logger
 
-**Default** `Logger.new(STDERR)`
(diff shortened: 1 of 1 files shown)


jessewgibbs:
@f3ndot we're hosting docs in a separate site now, so I went ahead and submitted the change there for review. Thanks for the contribution!

PR #740, opened 2018-05-17
Add Rollbar.configuration.logger_level

f3ndot:
This allows customization where Rollbar may be used inside a Rails project that has a production Rails.logger's log_level set to :info but does not want internal Rollbar entries all over the place.

Just because an app has its default logger set to a particular log level doesn't mean it wants Rollbar's own log level to mimic that.

### Realistic use case:

1. I use Rails
2. My `Rails.logger` goes to Logstash (`LogStashLogger`)
3. My log level is `:info` for valuable domain-specific investigation
4. I use Rollbar
6. By default Rollbar uses `Rails.logger`
6. Rollbar writes its internal logging to my logstash, reserved for domain-specific, terse info logs:

screen shot 2018-05-16 at 1 03 16 pm

7. Because of legacy reasons, there are a lot of workers or processes that generate Rollbar instances but don't want that polluting logstash

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,12 @@ end
 What logger to use for printing debugging and informational messages during
 operation.
 
+### logger_level
+
+**Default** `:info`
+
+Regardless of what Logger you're using, Rollbar will not proxy logs to it if its less than this particular level.
+
 ### disable_monkey_patch
 
 **Default** `false`
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -12,7 +12,7 @@ module Rollbar
     attr_accessor :custom_data_method
     attr_accessor :delayed_job_enabled
     attr_accessor :default_logger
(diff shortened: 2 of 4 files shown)


ArturMoczulski:
@f3ndot awesome stuff.  We have some test failures in the travis builds though. Do you want to check that out?

f3ndot:
Huh! I swear I resolved these locally. Working on it now

f3ndot:
@ArturMoczulski I simple logic expression mistake on my part 😉 we should be good to go now

f3ndot:
@ArturMoczulski some assistance with the remaining failed builds would be nice. Looks like the failures have nothing to do with my code changes now and, also, to have the majority of 1.9.x tests pass I had to use a notation that breaks codeclimate and code standards.

Please advise.

ArturMoczulski:
@f3ndot this looks good. Thank you for your contribution. Merging now.

PR #743, opened 2018-05-21
Update CHANGELOG for 2.16.0 release

prsimp:
While updating to 2.16.0, I noticed that the release info didn't make it into the CHANGELOG. This is just a reformatting of the [release notes](https://github.com/rollbar/rollbar-gem/releases/tag/v2.16.0) to fit the CHANGELOG's style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## 2.16.0
+- Anonymize IP address config option [#728](https://github.com/rollbar/rollbar-gem/issues/728)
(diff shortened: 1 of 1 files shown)


rivkahstandig3636:
Hi there, We're actually moving away from the Changelog to release notes.

prsimp:
> Hi there, We're actually moving away from the Changelog to release notes.

Ahh that explains it! In that case, I'd suggest dropping a note at the top of the CHANGELOG to that effect.

Thanks for all the work on Rollbar & this gem :beers:

rivkahstandig3636:
Good idea!

PR #744, opened 2018-05-21
Fix shoryuken handler undefined method call

herr-zimt:
[`process_async_item(item)`](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/notifier.rb#L668) tries to call `configuration.async_handler.call(item.payload)` for `Rollbar::Delay::Shoryuken` class even though there's no `self.call` defined which results in
```ruby
#<NoMethodError: undefined method `call' for Rollbar::Delay::Shoryuken:Class>
```


herr-zimt:
Yes, let's close this one.

PR #757, opened 2018-06-25
Support URI reserved characters in proxy server passwords.

petelacey:
I was the guy who submitted the original PR for proxy server support in Rollbar. Unfortunately, that code will fail in the proxy requires authentication, the credentials include URI-reserved characters, such as '@' or '?', and the credentials are provided as part of the Rollbar configuration.

This PR addresses that by holding the proxy configuration information in a Struct instead of a URI object. This PR includes tests, but the fix has also been tested in a runtime environment.

### DIFF
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -530,14 +530,15 @@ module Rollbar
     end
 
     def proxy_from_config
-      proxy = configuration.proxy
-      return nil unless proxy
+      proxy_settings = configuration.proxy
+      return nil unless proxy_settings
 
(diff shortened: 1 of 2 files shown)


rokob:
This is an interesting case, Net::HTTP is just fine with proxy passwords with special characters but URI is not. I just double checked that sending a request with Net::HTTP after setting the proxy password to something with an `@` works just fine. `URI#password=` on the other hand throws a `URI::InvalidComponentError: bad component(expected user component)`. This seems like a reasonable approach to me because we never use the returned URI object as anything but a bag of properties.

PR #730, opened 2018-04-27
update `append_nounce?` method for SecureHeaders

marunbai:
This pr try to solve this issue: #695

According to http://www.rubydoc.info/gems/secure_headers/SecureHeaders/Configuration#default-class_method, the get method is returning default config without any arg so we could just use default instead of get

PR #776, opened 2018-09-07
Update Rollbar gem to support secure_headers 6.x

agibralter:
The secure_headers 6.x gem changed the interface for accessing the
configuration object. Trying to use `default` yields an exception now.
This is the new interface.

I know that this was _just_ changed for @marunbai's PR https://github.com/rollbar/rollbar-gem/pull/730. However, it looks like the latest version of secure headers has changed this yet again. This fixes it immediately.

A few additional thoughts:

1) Since there is no hard dependency on a specific version of SecureHeaders in the Gemfile, it makes these kinds of changes really hard to track—they just pop up on you and break things... would it make sense to introduce a hard dependency on SecureHeaders? Probably not as that is probably not something many people want.

2) ... so, would it make sense to pull some of this coupling between rollbar-gem and secure-headers into a separate integration gem, say, `rollbar-gem-secure-headers` so that dependencies could be more accurately tracked?

I'd love to get _this_ fix out asap, but it might be worth thinking about 2...

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -157,8 +157,8 @@ module Rollbar
       def append_nonce?
         defined?(::SecureHeaders) && ::SecureHeaders.respond_to?(:content_security_policy_script_nonce) &&
           defined?(::SecureHeaders::Configuration) &&
-          !::SecureHeaders::Configuration.default.csp.opt_out? &&
-          !::SecureHeaders::Configuration.default.current_csp[:script_src].to_a.include?("'unsafe-inline'")
+          !::SecureHeaders::Configuration.dup.csp.opt_out? &&
+          !::SecureHeaders::Configuration.dup.csp[:script_src].to_a.include?("'unsafe-inline'")
       end
     end
   end
--- a/spec/rollbar/middleware/js_spec.rb
+++ b/spec/rollbar/middleware/js_spec.rb
@@ -177,7 +177,7 @@ END
           Object.const_set('SecureHeaders', Module.new)
           SecureHeaders.const_set('VERSION', '3.0.0')
           SecureHeaders.const_set('Configuration', Module.new {
-            def self.default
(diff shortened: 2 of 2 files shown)


rivkahstandig3636:
@ArturMoczulski or @jondeandres Could one of you review this?

jondeandres:
@agibralter thanks for this. I had started working on the code to make the secure_headers integration to work with, almost all the versions we can find. I don't think there should be a hard dependency with a such specific gem like secure_headers. Many people would complain about it.

So...I will try to work in an integration that will work with the different secure headers versions. Until we release it you could easily monkey patch `Rollbar::Middleware::Js` with the code you purpose in this PR.

I don't close the PR since I might use the commit on it or I will close after we release a final fix for this (that will be in next days)

jondeandres:
hey @agibralter, for those projects not using secure headers the unique overhead will be ` return SecureHeadersFalse unless defined?(::SecureHeaders::Configuration)` per request, I think that will be ok.

About extracting the logic to a gem, I think we can just write some code that will identify how to get the CSP from secure_headers configuration for the different secure_headers versions. We shouldn't force, if possible, to use one specific version of the gem.

agibralter:
@jondeandres I'm a little surprised to hear you're so resistant to extracting secure_header-related code into a separate gem. How do you plan to support all past and future versions of secure_headers? That is the whole point of gemspec and semantic versioning: when you depend on a third party library, you specify which versions of that library you support. The current plan is very likely to break—without warning—if/when secure_headers 7.x changes the API again.

jondeandres:
This is fixed on https://github.com/rollbar/rollbar-gem/pull/779

agibralter:
@jondeandres thank you! Are you planning on cutting a new version for rubygems soon?

ArturMoczulski:
Hey @agibralter . Yes, I will be cutting a new release either today or tomorrow.

Issue #712, opened 2018-04-14
Rollbar.js doesn't work with SecureHeaders


koss-lebedev:
Hello guys,

I found that Rollbar.js seems to have issues loading when `SecureHeaders` gem adds CSP policies. I see this error in logs:

`[Rollbar] Rollbar.js could not be added because undefined method 'current_csp' for #<SecureHeaders::Configuration:0x00000006bccfe0> exception`

 Is there a CSP rule that has to be added, or is this a known bug? Thank you

Full details: 
```
/home/deployer/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/logger.rb:452:in `error'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/logger_proxy.rb:28:in `log'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/logger_proxy.rb:22:in `error'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/notifier.rb:289:in `block (2 levels) in <class:Notifier>'
/home/deployer/.rvm/rubies/ruby-2.2.2/lib/ruby/2.2.0/forwardable.rb:183:in `log_error'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/middleware/js.rb:68:in `rescue in add_js'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/middleware/js.rb:58:in `add_js'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/rollbar-2.13.3/lib/rollbar/middleware/js.rb:25:in `call'
/home/deployer/servable/shared/bundle/ruby/2.2.0/bundler/gems/griddler-ses-888cee904cc6/lib/griddler/ses/middleware.rb:16:in `call'
/home/deployer/servable/shared/bundle/ruby/2.2.0/bundler/gems/griddler-ses-888cee904cc6/lib/griddler/ses/middleware.rb:16:in `call'
/home/deployer/servable/shared/bundle/ruby/2.2.0/gems/twilio-ruby-4.11.1/lib/rack/twilio_webhook_authentication.rb:28:in `call'
[...]

PR #779, opened 2018-09-11
Support multiple secure_headers versions

jondeandres:
This PR tries to fix https://github.com/rollbar/rollbar-gem/issues/712 and make the gem work with different secure_headers versions.

https://github.com/twitter/secure_headers doesn't define a `VERSION` constant and makes tricky to detect which interface to use. Depending on few checks we return a class that will take care of the interface to use.

I've tested manually the implementation using secure_headers versions from 2.x to 6.x passing by different versions 3.x (which is the major version with more minor versions). Things seem to work fine and now looks a bit easier extending the behavior if secure_headers changes the interface...

Some mocks are used in the tests to simulate the different secure_headers versions. Tests have been refactored a bit so we have few shared examples that can be used from secure_headers > 3.x.

It seems trying to be compatible with secure_headers 2.x is not easy cause the way they define the `nonce` so that version will be still unsupported.

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -36,6 +36,8 @@ module Rollbar
         end
       end
 
+      private
+
       def enabled?
         !!config[:enabled]
       end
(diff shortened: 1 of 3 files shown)


ArturMoczulski:
Looks great. Thanks @jondeandres

PR #785, opened 2018-10-29
Test Ruby 2.3 and Rails 5.2 in CI.

boone:
The CI for this gem is behind on the latest Rubies and Rails.

There are three commits:

* Add Ruby 2.3.0.
* Skip JRuby 1.8 tests for Rails 5.
* Add a Rails 5.2 Gemfile and test Ruby 2.2.2 and 2.3.0 against it.

I'd like to add Ruby 2.4 and 2.5 as well, but these versions seem to have some issues with the current gemfiles. I'll look into those later if I have time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ gemfile:
   - gemfiles/rails42.gemfile
   - gemfiles/rails50.gemfile
   - gemfiles/rails51.gemfile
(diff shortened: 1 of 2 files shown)


jessewgibbs:
@boone thanks for the contribution.

ArturMoczulski:
Looks grat @boone . Merged into `master`.

PR #786, opened 2018-10-30
Make `logger` a public method.

boone:
In Ruby 2.4+, Rollbar will emit a warning:

`forwarding to private method Rollbar::Notifier#logger`

The delegators of the Ruby Forwardable module expect that all the methods be public. Rollbar has them defined as `PUBLIC_NOTIFIER_METHODS` [here](https://github.com/boone/rollbar-gem/blob/master/lib/rollbar.rb#L25-L33), but the `logger` method is currently private.

This PR just moves the logger method to the public section of the `Rollbar::Notifier` class.

### DIFF
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -295,6 +295,10 @@ module Rollbar
       end
     end
 
+    def logger
+      @logger ||= LoggerProxy.new(configuration.logger)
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
Awesome PR. Thanks @boone . Merged into `master`.

PR #788, opened 2018-10-31
Rework gem regex spec to avoid deprecation warning.

boone:
Currently the specs are outputting this warning:

```
NOTE: Dependency.new w/ a regexp is deprecated.
Dependency.new called from .../lib/ruby/2.2.0/rubygems/specification.rb:910
```

In #120, the warning was fixed in the main code but not the specs. This PR fixes the specs using the same lookup.

This will also help future-proof the specs, as calling `Gem::Specification.find_all_by_name` with a regex in Ruby 2.4+ will raise an exception:

```
NoMethodError: undefined method `upcase' for /rspec/:Regexp
```

### DIFF
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -1420,7 +1420,10 @@ describe Rollbar do
         config.project_gems = gems
       end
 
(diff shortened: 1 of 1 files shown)


jessewgibbs:
@boone thanks for the contribution.

ArturMoczulski:
Looks great @boone thanks for PR

PR #789, opened 2018-11-01
Update gemspec to not include test files to reduce gem byte size

gabetax:
The rollbar .gem file is [currently 264kb](https://rubygems.org/gems/rollbar/versions). The majority of this is from unnecessarily including the 800kb (uncompressed) spec/ files in the gem, which is unnecessary. For the `files` directive, I've replicated the current syntax from the current `bundle gem foo` generator. I have also removed the `test_files` directive, which doesn't seem useful, nor is it in [the current gemspec reference](https://guides.rubygems.org/specification-reference/).

I've manually confirmed that `gem build rollbar.gemspec` succeeds with the changes.

Pros: Less network bandwidth, less disk consumed, slightly faster builds. Cons: None.

### DIFF
--- a/rollbar.gemspec
+++ b/rollbar.gemspec
@@ -12,8 +12,9 @@ Gem::Specification.new do |gem|
   gem.homepage      = 'https://rollbar.com'
   gem.license       = 'MIT'
 
-  gem.files         = `git ls-files`.split($\)
-  gem.test_files    = gem.files.grep(%r{^(spec)/})
+  gem.files         = Dir.chdir(File.expand_path('..', __FILE__)) do
+    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  end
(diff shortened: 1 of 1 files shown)


jessewgibbs:
@gabetax thanks for the contribution!

gabetax:
Opinions on handling Style/ExpandPathArguments? `__dir__` is not compatible with the jruby 1.9 tests. Should the .rubocop.yml specify 1.9 as its target version? I can also just throw away the (not particularly useful) chdir and defer this problem.

gabetax:
ping @ArturMoczulski 

I gave up on the Dir.chdir, since I don't think it's possible for it to be green under both jruby and the rubocop assertions (and it's not extremely valuable)

ArturMoczulski:
This looks like a great improvement. Thanks @gabetax for the PR. Merging into `master` now.

PR #792, opened 2018-11-29
Change the meaning of scrub_whitelist

rokob:
scrub_whitelist should be a list of keys that we do not want to scrub even if they are in scrub_fields

### DIFF
--- a/spec/rollbar/scrubbers/params_spec.rb
+++ b/spec/rollbar/scrubbers/params_spec.rb
@@ -311,7 +311,7 @@ describe Rollbar::Scrubbers::Params do
         [:secret, :password]
       end
       
-      let(:whitelist) { true }
+      let(:whitelist) { [:password] }
 
       context 'with Array object' do
         let(:params) do
@@ -327,8 +327,8 @@ describe Rollbar::Scrubbers::Params do
         let(:result) do
           [
             {
-              :foo => /\*+/,
-              :secret => 'the-secret',
+              :foo => 'bar',
+              :secret => /\*+/,
               :password => 'the-password',
               :password_confirmation => 'the-password'
             }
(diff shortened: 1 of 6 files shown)


brianr:
What's the backwards-compatibility situation for this change? i.e. how will this impact customers who already have this setting in their config but then upgrade to the newest version?

rokob:
It was introduced in 2.16.4 which was two releases ago. We don't want that behavior so I think making a breaking change over this short of a time period is a better choice than supporting the version that took a boolean. We can probably just bump the minor version, we don't have to go to 3.0 as long as it is clear what is going on in the release notes. I wouldn't expect a lot of people to be using this feature yet.

brianr:
Sounds good 👍

spec/rollbar/scrubbers/params_spec.rb
```diff
@@ -327,8 +327,8 @@
         let(:result) do
           [
             {
-              :foo => /\*+/,
-              :secret => 'the-secret',
+              :foo => 'bar',
+              :secret => /\*+/,
               :password => 'the-password',
               :password_confirmation => 'the-password'
```
tjallingvanderwal: I would expect a whitelist to only allow exact matches. Not substring/fuzzy matches.
So `scrub_whitelist = [:password]` should result in `:password_confirmation` being scrubbed.

rokob: Yeah that makes sense, it is a byproduct of the current way scrub_fields end up doing substring matches. The fact is if you had `password_confirmation` in the scrub_fields and put `password` in the whitelist then `password_confirmation` would get scrubbed. If you only have `password` in the scrub_fields, then both `password` and `password_confirmation` end up getting scrubbed due to the substring match, but then neither get scrubbed if you whitelist `password`.

ArturMoczulski:
LGTM. I'm merging into `master`.

Issue #793, opened 2018-12-05
Whitelisting doesn't work (for me)


gbrylinski:
### Before #792 

Before this patch, whitelisting was working properly except situation when `config.filter_parameters` was set for Rails application. `config.filter_parameters` is supposed to list fields that are *blacklisted* and should be excluded/scrubbed from logs, reports, etc. Pre #792 code adds content of `config.filter_parameters` to blacklist as well as whitelist what can lead to passing sensitive data into report.

so, for configuration like this:

```ruby
# application.rb
config.filter_parameters = [:password, :api_key]

# initializers/rollbar.rb
config.scrub_whitelist = true
config.scrub_fields = [:user_id, :name, :surname]
```

I would expect that only fields `[:user_id, :name, :surname]` keep real values and any other field is scrubbed. But, because `[:password, :api_key]` are added to the list also `password` and `api_key` values are whitelisted and get reported.

### After #792 

it seems to not work. I started with following config:

```ruby
# initializers/rollbar.rb
config.scrub_whitelist = [:password]
```

I would expect all fields except `password` to be scrubbed but all values are preserved. For:

```ruby
# initializers/rollbar.rb
config.scrub_whitelist = [:password]
config.scrub_fields = []
```

again, all except `password` should get scrubbed but again all are preserved. When I use:

```ruby
# application.rb
config.filter_parameters = [:utf8, :commit]

# initializers/rollbar.rb
config.scrub_whitelist = [:password]
config.scrub_fields = [:username]
```

I get `utf8`, `commit` & `username` scrubbed and everything else is whitelisted. It seems like whitelist is applied to blacklist and then blacklisted fields get scrubbed. My understanding of whitelisting is slightly different - we let pass only things that are on the whitelist and anything else get scrubbed. That's also what documentation says:

```
scrub_whitelist

Set to true to switch scrubbing mode to whitelist. In whitelist mode, everything but the fields provided in scrub_fields gets scrubbed. The values will be replaced with a "*" before being sent.
```

### Pull request #794

I've made a PR based on master before #792 was applied. It does following things:

- changes specs so they test whitelist behaviour instead of blacklist
- changes `Url` & `Params` scrubbers
- makes necessary changes to `RequestDataExtractor` to support whitelist in urls
- properly ignores `filter_parameters` when whitelisting is used

rokob:
[...]

PR #794, opened 2018-12-05
Make whitelisting ignore filter_parameters

gbrylinski:
After testing #792 I think it's not working properly. This fix is based on previous master that worked well except situation when `config.filter_parameters` was set for Rails app. 

Detailed explanation can be found in the issue #793.

It is based on current documentation so to use whitelisting following config should be used:

```ruby
config.scrub_whitelist = true
config.scrub_fields = [:username, :action, :id, :something_else]
```

### DIFF
--- a/lib/rollbar/request_data_extractor.rb
+++ b/lib/rollbar/request_data_extractor.rb
@@ -60,7 +60,8 @@ module Rollbar
     def scrub_url(url, sensitive_params)
       options = {
         :url => url,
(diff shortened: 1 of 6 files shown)


jessewgibbs:
@rokob or @ArturMoczulski - could you guys take a look at this?

rokob:
I am going to close this in favor of https://github.com/rollbar/rollbar-gem/pull/795 which should make it possible to do what you want correctly, albeit not with the exact configuration you describe above. `scrub_whitelist` as a boolean is not what we want and the docs will be updated to reflect that at the next release.

PR #795, opened 2018-12-05
Make scrub_all work correctly (including with whitelist)

rokob:
`:scrub_all` as a key in the `scrub_fields` configuration means to scrub everything. This kinda sorta worked before, but there were some corner cases that were broken. This fixes those.

Also, `scrub_whitelist` is intended to be a list of things you do not want to scrub regardless of why you were going to scrub them in the first place. This was broken with `scrub_all` before, this fixes that.

Furthermore, the `scrub_whitelist` does an exact match, so if you for example have `:password` in `scrub_fields`, because we do substring matching for scrubbing, this would imply `password` and `password_confirmation` would get scrubbed. If you decide to add `password` to `scrub_whitelist`, you will see that `password` will not be scrubbed but `password_confirmation` will still be scrubbed. This is because the substring matching for scrubbing is still matching `password_confirmation` but the whitelist requires an explicit match which does not happen with only `password` in the whitelist. This seems to be the right behaviour as a closed default is what one wants when it comes to sensitive data.

### DIFF
--- a/lib/rollbar/scrubbers/params.rb
+++ b/lib/rollbar/scrubbers/params.rb
@@ -6,7 +6,8 @@ module Rollbar
     # This class contains the logic to scrub the received parameters. It will
     # scrub the parameters matching Rollbar.configuration.scrub_fields Array.
     # Also, if that configuration option is set to :scrub_all, it will scrub all
-    # received parameters
+    # received parameters. It will not scrub anything that is in the scrub_whitelist
+    # configuration array even if :scrub_all is true.
     class Params
       SKIPPED_CLASSES = [::Tempfile]
       ATTACHMENT_CLASSES = %w(ActionDispatch::Http::UploadedFile Rack::Multipart::UploadedFile).freeze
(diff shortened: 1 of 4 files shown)


ArturMoczulski:
LGTM. @rokob is there anything that needs to be added here or is this ready for merge?

rokob:
It's ready to go.

domininik:
hi @ArturMoczulski, it seems you released wrong pull request (?).
release note of https://github.com/rollbar/rollbar-gem/releases/tag/v2.18.1 links this pull request but the one which was actually merged was https://github.com/rollbar/rollbar-gem/pull/792

here https://github.com/rollbar/rollbar-gem/issues/793 we discussed why https://github.com/rollbar/rollbar-gem/pull/792 is not enough and there was https://github.com/rollbar/rollbar-gem/pull/794 which has been replaced by https://github.com/rollbar/rollbar-gem/pull/795 :)

ArturMoczulski:
My bad. Fixing this now.

PR #799, opened 2019-01-16
Tune up RSpec and Travis

waltjones:
- [x] Get all the current environments green
- [x] Enable Ruby 2.4.x, 2.5.x, 2.6.x
- [x] See what else comes up

This PR brings RSpec and Travis status up to:
* All MRI are green except 4.2.x. This includes: 1.8.7, 1.9.2, 1.9.3, 2.0.0, 2.1.0, 2.2.0, 2.3.0, 2.3.8 (last 2.3.x), 2.5.3 (last 2.5.x), and 6.0.0. These all work for me locally under rvm as well.
* Travis JRubies 18mode and 19mode are enabled, but have dependency errors. I could not find a prior build where these worked. Same with Rubinius.
* Travis has, apparently very recently, removed openjdk6 and oraclejdk7 from the Trusty distribution. (Other distros have even less stuff. Trusty is the default.) The config has been updated to use openjdk7, openjdk8, oraclejdk8, and oraclejdk9. While Jruby still has issues on these builds, each of the JDK environments installs correctly.

There is a matrix outlining the status of each of the Ruby and Rails combinations, with a description of known issues. As soon as any of the JRubies are working they will be added. This matrix may move, and if so I’ll update this comment.
https://docs.google.com/spreadsheets/d/1keJNwA2vHvyCZHKilvcB2JnPRzlvoiZBUHy9lK2ppkk/edit#gid=0

Managing the different environments is made significantly more complex because Travis appears to only allow a single global matrix for configuration. It would be much more manageable if multiple matrices could be defined, each with its own environment. (Anyone that knows more about how to do this, or if it’s possible, please reach out.)

The rubygems version must be set at the max version required by all rubies. Similarly, the bundler version must be set at a version compatible with all rubies. The current settings in .travis.yml are working. But Jruby and Rubinius builds aren’t enabled yet, and the current blocking issue is related to versions and dependencies. Having to rely on a single global version for these dependencies seems very brittle.

There is a tricky cutoff at Ruby 2.4.x and Rails 4.2.x where Ruby aliases Fixnum and Bignum to a single class, Integer. Versions of Rails that do not handle this correctly can (and do) stack overflow, giving either a system segfault or a more polite Stack Too Deep error. This is easily reproducible both locally and in Travis, and is noted in the linked matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: ruby
 before_install:
   - gem update --system 2.5.1
   - gem --version
-  - gem install bundler
+  - gem install bundler -v 1.6.1
   - bundle --version
 rvm:
   - 1.9.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,24 +4,29 @@ services:
 language: ruby
 # Broken bundler on travis CI - https://github.com/bundler/bundler/issues/2784
 before_install:
-  - gem update --system 2.5.1
+  - gem update --system 2.7.7
   - gem --version
   - gem install bundler -v 1.6.1
   - bundle --version
+
 rvm:
   - 1.9.3
   - 2.0.0
   - 2.1.0
   - 2.2.2
   - 2.3.0
+  - 2.3.8
+  - 2.4.5
+  - 2.5.3
+  - 2.6.0
   - jruby-18mode
   - jruby-19mode
   - rbx
 jdk:
-  - openjdk6
   - openjdk7
-  - oraclejdk7
+  - openjdk8
   - oraclejdk8
+  - oraclejdk9
 env:
   - ROLLBAR_SSL_CERT_FILE=/home/travis/build/rollbar/rollbar-gem/spec/cacert.pem
 gemfile:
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
 language: ruby
 # Broken bundler on travis CI - https://github.com/bundler/bundler/issues/2784
 before_install:
-  - gem update --system 2.1.11
+  - gem update --system 2.5.1
```
waltjones: Some jruby builds were timing out on rubygems version.

lib/rollbar/delay/sidekiq.rb
```diff
@@ -10,7 +10,7 @@ def initialize(*args)
       end
 
       def call(payload)
-        if (::Sidekiq::Client.push @options.merge('args' => [payload]) == nil)
+        if ::Sidekiq::Client.push(@options.merge('args' => [payload])) == nil
```
waltjones: Previous code was sending the result of the `==` expression to `#push` instead of the options hash.

spec/delay/sidekiq_spec.rb
```diff
@@ -35,8 +35,8 @@ class Sidekiq
 
     context "with default options" do
       it "enqueues to default queue" do
-        options = Rollbar::Delay::Sidekiq::OPTIONS.merge('args' => payload)
-        ::Sidekiq::Client.should_receive(:push).with options
+        options = Rollbar::Delay::Sidekiq::OPTIONS.merge('args' => [payload])
+        ::Sidekiq::Client.should_receive(:push).with(options){ true }
```
waltjones: Since we are recently now using the return value of `::Sidekiq::Client.push`, this needs to mock a return value.

ArturMoczulski:
@waltjones looks great. I'm merging this in.

lib/rollbar/delay/sidekiq.rb
rokob: Good catch, this is crazy, I was suddenly worried that this has somehow been here for a long time but just saw that the conditional is new enough not to have been released yet.

ArturMoczulski: @waltjones thanks for catching that

Issue #782, opened 2018-10-03
Update Deploy API wrapper to support status property


jessewgibbs:
See https://docs.rollbar.com/v1.0.0/reference#post-deploy for details of the new options for this API.

ArturMoczulski:
Implemented and merged into `master`

PR #800, opened 2019-01-17
GitHub Issue #782: Update Deploy API wrapper to support status property

ArturMoczulski:
Add support for the `status` field for Deploy Tracking. This required not only including the field in the API request but also changing the flow of the `rollbar:deploy*` tasks.

In addition, this PR removes code duplication between Capistrano 2 and 3 logic for the most part.


ArturMoczulski:
I'm going to take care of the Codeclimate and Codacy warnings. Other than that @rokob @waltjones can you guys take a look at this PR?

waltjones:
Looking now.

waltjones:
Do we, or where do we test the capistrano integration? My global grep of the project didn't turn up anything.

ArturMoczulski:
We do not have Capistrano tests. I believe building a test suite for Capistrano tasks is a pretty big chunk of work in itself so I did not include it in this batch of work.

The only part that we could potentially add tests for in here now is `::Rollbar::Deploy` module however those would be more of integration tests against the Rollbar API which from what I understand we don't anywhere really. So I didn't go into that.

waltjones:
Makes sense.

lib/rollbar/capistrano_tasks.rb
```diff
@@ -0,0 +1,131 @@
...
+              else
+                logger.error 'Unable to update deploy status in Rollbar'
+              end
+            end
+          end
+        end
+      end
+
+      def capistrano_300_warning(logger)
+        logger.warn("You need to upgrade capistrano to '>= 3.1' version in order to correctly report deploys to Rollbar. (On 3.0, the reported revision will be incorrect.)") if ::Capistrano::VERSION =~ /^3\.0/
```
nisanth074, line 66: @ArturMoczulski Capistrano 2 (atleast version 2.12.0) doesn't have the `Capistrano::VERSION` constant.

Does the latest version of Rollbar not support Capistrano 2?

Issue #583, opened 2017-04-08
Missing coverage in Codacy


coryvirok:
https://codeclimate.com/github/rollbar/rollbar-gem

Looks like @jondeandres had this working at one point or at least tried to get it working.

coryvirok:
@jondeandres any thoughts on getting test coverage reporting to CodeClimate?

jondeandres:
I had enabled it on our spec_helper.rb file, https://github.com/rollbar/rollbar-gem/blob/master/spec/spec_helper.rb#L1-L14, but the required gems are not present in our gemfiles, so that may be the reason because https://codeclimate.com/repos/537e61f8e30ba05e65002eaa/settings/test_reporter is empty.

I will take a look at this.

coryvirok:
I haven't been very happy with CodeClimate but Codacy seems great. Let's set up test coverage there. https://www.codacy.com/app/Rollbar/rollbar-gem/settings/coverage#IntegrationsTab

rivkahstandig3636:
Hi there, I’m closing out all issues opened before 2018 that haven’t had any activity on them since the start of this year. If this is still an issue for you, please comment here and we can reopen this. Thanks!

coryvirok:
Reopening so that we can get coverage info reported to Codacy.

waltjones:
With the linked PR, this is working now.

PR #804, opened 2019-01-19
Issue #583: Enable Codecy coverage with simplecov

waltjones:
Issue https://github.com/rollbar/rollbar-gem/issues/583

Coverage results are available under the Codacy dashboard "Files" tab. Results don't yet appear under PR results, which I think is because it relies on having data for master. Travis builds are confirmed to be sending data successfully. If PRs don't start showing results after merge, I'll ask Codacy about it.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,15 +1,19 @@
 begin
   require 'simplecov'
-  require 'codeclimate-test-reporter'
+  require 'codacy-coverage'
+
+  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
+    SimpleCov::Formatter::HTMLFormatter,
+    Codacy::Formatter
+  ]
 
   SimpleCov.start do
     add_filter '/spec/'
-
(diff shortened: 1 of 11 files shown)


spec/spec_helper.rb
```diff
@@ -1,15 +1,19 @@
...
-
-    formatter SimpleCov::Formatter::MultiFormatter.new([
-      SimpleCov::Formatter::HTMLFormatter,
-      CodeClimate::TestReporter::Formatter
-    ])
   end
+
+  # Skip Codacy when running locally, to display the Simplecov summary in the console
+  # and write an updated coverage/index.html
+  Codacy::Reporter.start if Codacy::Formatter.new.send :should_run?
```
waltjones: Codacy's `should_run?` logic detects whether it is in a local or CI environment, and Codacy does not send data when local. Unfortunately, it does still block Simplecov from writing its local output. So here we ensure Codacy doesn't interfere with Simplecov when running local.

Issue #765, opened 2018-07-23
Session key should not be send to rollbar by default


grosser:
Atm rollbar sends the session id which can be replayed to log in as every user that ever had an error 😱 

`Samson::Application.config.session_options.fetch(:key)`
... but that won't work since rollbar initializer often comes before session (alphabetical) ... so idk how to fix that generally, but it's super scary imo ...

fixed this for our app here https://github.com/zendesk/samson/pull/2861 by defining the key globally and than adding it to scrub_fields

ArturMoczulski:
@grosser can this be closed now?

grosser:
should be fixed in rollbar since I'd assume almost every app using rollbar will send each users session-ids to rollbar which means anyone with access to rollbar can login as each user

ArturMoczulski:
This is a conversation we should bring @brianr @coryvirok and @rokob to this conversation. I believe our approach with all SDKs right now is to send all of the session data unless you configure the SDK otherwise. I agree with you @grosser that we should probably scrub that by default, but I need a broader view to see what impact is that going to have on the utility of error reports.

I believe session ID would be considered personally identifiable information in the light of the GDPR here actually, but I'm not 100% certain.

jessewgibbs:
This sounds like something we should consider across all our SDKs.  I agree w/ @ArturMoczulski that session-id should probably be filtered by default, especially in light of GDPR.

grosser:
I'd say GDPR is the smaller issue, having something in the logs that any
person with access to rollbar can use to log in as any user in the app is
super scary.

PR #806, opened 2019-01-21
GitHub Issue #765: Session key should not be send to rollbar by default

ArturMoczulski:
Add scrubbing `session_id` from payload by default. Requested in https://github.com/rollbar/rollbar-gem/issues/765


ArturMoczulski:
https://docs.rollbar.com/docs/gem-configuration-reference needs to be updated after this is released

Issue #698, opened 2018-03-08
(request) Support Rails 5.2 Content Security Policy


brian-kephart:
The current implementation for using rollbar.js with this gem uses an inline script, which requires that the CSP (if any) allows 'unsafe-inline' scripts.

[Looking through the code](https://github.com/rollbar/rollbar-gem/blob/f4c1afc9e2b269e9e5298440581e8b76f6cb5ffa/lib/rollbar/middleware/js.rb#L141), it appears that this issue is already solved using a nonce for users of the secure-headers gem, but no such fix is applied for the new built-in CSP feature of Rails 5.2.

rivkahstandig3636:
Thanks for the report!

waltjones:
With the linked PR, this is available on master, and in the next release.

PR #808, opened 2019-01-22
WIP: Enable Rails 5 content_security_policy for js snippet

waltjones:
Issue https://github.com/rollbar/rollbar-gem/issues/698

Rails 5.2 has built in support for CSP: https://edgeguides.rubyonrails.org/security.html#content-security-policy

Apps that want to enable CSP have two choices for inline JS, like the rollbar js snippet:
1. Set the `'unsafe-inline'` policy, which is global and removes security for inline JS, or
2. Use a nonce, which Rails CSP supports, but requires the snippet to use the nonce.

This PR supports both options by only including the nonce when expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,36 @@ rvm:
   - 2.4.5
   - 2.5.3
   - 2.6.0
-  - jruby-18mode
-  - jruby-19mode
   - rbx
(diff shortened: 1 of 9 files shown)


waltjones:
Opening new PR without merge commit clutter.

Issue #759, opened 2018-06-27
Active Job adapter for delivering errors


morgoth:
Would it be possible to add support for AJ adapter for delivering errors?
It's missing now https://github.com/rollbar/rollbar-gem/tree/master/lib/rollbar/delay

I think it would be a nice addition for Rails apps

ArturMoczulski:
@morgoth I believe that's available as a plugin here https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/plugins/active_job.rb

All you have to do is include it in your `ActiveJob` implementation like so:

app/jobs/application_job.rb
```ruby
class ApplicationJob < ActiveJob::Base
  include Rollbar::ActiveJob
end
```

Let me know if this helps. If you have further questions, please comment and I'll reopen the issue.

morgoth:
@ArturMoczulski I was thinking about a way to deliver errors by AJ and not reporting errors in AJ.
So having configuration option, something like:
`config.use_active_job` that would use any background job backend that is configured in AJ

ArturMoczulski:
@morgoth this is available already. Here is how you enable reporting through DJ:
`config.use_delayed_job`

There are also other job handlers:
`config.use_sidekiq`
`config.use_resque`
`config.use_shoryuken`
`config.use_thread`
`config.use_sucker_punch`

Enabling any of the above will result in reporting through a background job backend.

Does this answer your question?

ArturMoczulski:
I see. So you mean, you would like us to support ActiveJob as an abstract and not necessarily a specific implementation, is that correct? So that you don't have to reconfigure Rollbar at all and only rely on what is configured in Rails as the AJ implementation. Did I get that right?

morgoth:
Yes, exactly :)

PR #810, opened 2019-01-24
Github Issue #759: Active Job adapter for delivering errors

ArturMoczulski:
This a PR fulfilling a request in https://github.com/rollbar/rollbar-gem/issues/759

It adds `use_active_jobs` config option which instructs the SDK to process messages asynchronously through the abstract ActiveJob framework and fully relies on it to process the report as a background job with whatever job processing backend is configured in Rails.

Thanks to this, users don't have to change their Rollbar configuration and app depending on what jobs processing backend is currently being used.

This is different to our other `use_*` config options which are enabling specific job processing backends implementations.

TODO: Update https://docs.rollbar.com/docs/gem-configuration-reference when this is released


deepfryed:
@ArturMoczulski Apologies for adding a comment / question here.

> Thanks to this, users don't have to change their Rollbar configuration and app depending on what jobs processing backend is currently being used.

This confused me a bit - are we able to use this somehow to have a single sidekiq worker that is able to asynchronously report rollbar errors from different Rails applications ? I can't see how that's possible, but it'd be good if we could do that.

Issue #696, opened 2018-03-06
ActionController::RoutingError (No route matches [GET] "/verify"):


damienlethiec:
Hello,

I often add rollbar to my projects without problems but this time it is not working:

Rails 5.2.0.rc1 (I have already been able to make Rollbar work in another rails 5.2.0.rc1 project).

When I run `rake rollbar:test`, I have an error (bellow).
```
Running via Spring preloader in process 85529
Testing manual report...
Setting up the controller.
Processing...
Started GET "/verify" for  at 2018-03-06 18:19:26 +0700
Cannot render console from ! Allowed networks: 127.0.0.1, ::1, 127.0.0.0/127.255.255.255
  
ActionController::RoutingError (No route matches [GET] "/verify"):
  
actionpack (5.2.0.rc1) lib/action_dispatch/middleware/debug_exceptions.rb:65:in `call'
rollbar (2.15.5) lib/rollbar/middleware/rails/show_exceptions.rb:22:in `call_with_rollbar'
web-console (3.5.1) lib/web_console/middleware.rb:135:in `call_app'
web-console (3.5.1) lib/web_console/middleware.rb:20:in `block in call'
web-console (3.5.1) lib/web_console/middleware.rb:18:in `catch'
web-console (3.5.1) lib/web_console/middleware.rb:18:in `call'
actionpack (5.2.0.rc1) lib/action_dispatch/middleware/show_exceptions.rb:33:in `call'
railties (5.2.0.rc1) lib/rails/rack/logger.rb:38:in `call_app'
railties (5.2.0.rc1) lib/rails/rack/logger.rb:26:in `block in call'
activesupport (5.2.0.rc1) lib/active_support/tagged_logging.rb:71:in `block in tagged'
activesupport (5.2.0.rc1) lib/active_support/tagged_logging.rb:28:in `tagged'
activesupport (5.2.0.rc1) lib/active_support/tagged_logging.rb:71:in `tagged'
railties (5.2.0.rc1) lib/rails/rack/logger.rb:26:in `call'
sprockets-rails (3.2.1) lib/sprockets/rails/quiet_assets.rb:13:in `call'
actionpack (5.2.0.rc1) lib/action_dispatch/middleware/remote_ip.rb:81:in `call'
actionpack (5.2.0.rc1) lib/action_dispatch/middleware/request_id.rb:27:in `call'
rack (2.0.4) lib/rack/method_override.rb:22:in `call'
[...]

PR #816, opened 2019-02-02
Issue #696: Fix rake rollbar:test and update for rubocop and rspec

waltjones:
https://github.com/rollbar/rollbar-gem/issues/696

Refactor rake rollbar:test:
- [x] Refactors rake_tasks.rb to use a module callable from Rspec
- [x] Refactors rake_tasks.rb to pass rubocop
- [x] Adds RSpec examples for primary use cases

Fix https://github.com/rollbar/rollbar-gem/issues/696:
- [x] Run the verification test using a clone of the default Rails app.
- [x] Resolve RSpec issues with Rails 4.x

The root cause of the issue, "ActionController::RoutingError (No route matches [GET] "/verify")" when running `rake rollbar:test:`, is that when Spring is running in the background and `config.cache_classes == false` in the current Rails environment, routes are reset on each request and the new test route can't be found.

The issue was initially difficult to reproduce because, even if Spring is already running, `bin/rake` will use Spring and reproduce the error. However, `bundle exec rake` does not use Spring and therefore does not reproduce the error.

In short, to see this error you must have Spring running, have `config.cache_classes == false`, and use `bin/rake` (or at least a way of invoking rake that does use Spring.)

Several techniques were considered to work around this issue. Neither the middleware chain (`ActionDispatch::Reloader` does the actual work resetting the routes) nor the `cache_classes` flag can be changed after Rails is loaded.

Cloning the app allows the config to be updated without modifying the default app. While changes to the default app config could be safely reverted by using `ensure` sections, the code is shorter and cleaner when using the cloned app.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ gem 'delayed_job', :require => false
 gem 'generator_spec'
 gem 'girl_friday', '>= 0.11.1'
 gem 'redis'
-gem 'resque'
+gem 'resque', '< 2.0.0'
 gem 'rspec-command'
 gem 'rubocop', :require => false
 gem 'sinatra'
--- a/gemfiles/rails30.gemfile
+++ b/gemfiles/rails30.gemfile
@@ -47,7 +47,6 @@ elsif RUBY_VERSION.start_with?('2')
 end
 
 gem 'sinatra'
-gem 'resque'
 gem 'delayed_job', :require => false
 gem 'redis'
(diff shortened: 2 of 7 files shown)


rokob:
I also looked at this earlier but apparently I didn't leave a comment. But yeah this looks great.

waltjones:
@ArturMoczulski , @rokob , thank you!

PR #822, opened 2019-02-14
Remove Ruby 1.8.7, 1.9.2 from Travis matrix, and update gemspec

waltjones:
Maintenance ended on Ruby 1.8.7 and 1.9.2 in 2014, including security updates. The latest Rails they work with is 3.x, and especially 1.8.7 doesn't support modern Ruby syntax, which holds back the gem code from using preferred syntax. Finally, in spite of their limited use and usefulness, it's not uncommon for a PR to need special treatment for these in the Travis matrix. RIP.

(Any users of these Ruby versions can still pin to an earlier version of rollbar-gem.)

Todo: Decide what version bump is appropriate for the change in gemspec requirements. (Ruby >= 1.9.3)
- [x] Update version.rb and readme.md with correct gem version.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -31,3 +31,6 @@ Metrics/BlockLength:
 
 Style/ExpandPathArguments:
   Enabled: false # syntax requires Ruby >= 2.0
+
+Gemspec/RequiredRubyVersion:
+  Enabled: false # rubocop compares to gemspec, yet won't allow 1.9 as minimum version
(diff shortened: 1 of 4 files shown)


ArturMoczulski:
@rokob @coryvirok how do you guys feel about dropping support for Ruby < 1.9.3 ? Do we have any users still using such legacy versions?

coryvirok:
I don't know the answer to your question, @ArturMoczulski. I would be *very* careful to not break backward compatibility in the gem though. If this is important enough, I think we can get the data but it'll take some time.

Issue #691, opened 2018-02-24
support cheaper ignore in config.before_process


grosser:
Would be nice if returning `:ignore` or something like that worked too so I don't have to raise and exception, which is kinda expensive and pollutes debug output

rivkahstandig3636:
Thanks for the suggestion @grosser!

ArturMoczulski:
@grosser I'm not sure I understand. You want to be able to trigger ignoring by returning `:ignore` instead of having to raise `Rollbar::Ignore`?

grosser:
yes, since that would be far cheaper (cpu/memory) than raising+catching an
exception

PR #824, opened 2019-02-14
Allow return 'ignored' in before_process handlers'

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/691

Allow before_process handlers to either raise `Rollbar::Ignore` as usual, or return `'ignored'`.

### DIFF
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -134,10 +134,11 @@ module Rollbar
       return 'ignored' if ignored?(exception, use_exception_level_filters)
 
       begin
-        call_before_process(:level => level,
-                            :exception => exception,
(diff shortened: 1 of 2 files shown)


grosser:
why a string and not `:ignored` ?

waltjones:
@grosser, Internal code in the same code path already uses the string `'ignored'` with the same semantic meaning, and also other return values as strings. Doing so was more consistent with existing code. Error levels, for example, are also represented as strings and are used in public interfaces such as `before_process`, `exception_level_filters`, etc.

I'd like to revisit at some point. Symbols would be best practice, but only makes sense if used generally and not just in one place.

grosser:
sounds good enough! :)

Issue #719, opened 2018-04-24
Add capture uncaught config option


rivkahstandig3636:
Similar to rollbar.js, the ability to turn on/off uncaught error reporting.

PR #826, opened 2019-02-19
Add capture_uncaught flag to config options

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/719

Set `capture_uncaught = false` to disable Rollbar handling of uncaught exceptions.

- [x] Add capture_uncaught behavior.
- [x] Refactor lib/rollbar/exception_reporter.rb for rubocop compliance.

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -8,6 +8,7 @@ module Rollbar
     attr_accessor :async_handler
     attr_accessor :branch
     attr_reader :before_process
+    attr_accessor :capture_uncaught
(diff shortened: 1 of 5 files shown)


rokob:
This means we still catch the exception, we just don't report it, right?

waltjones:
@rokob, yes, correct. All paths rescue, then reraise the exception (same as before, and same as without the flag.) Only the reporting method `report_exception_to_rollbar` is short circuited.

waltjones:
I've updated the rails rspec example to expect the exception. (The rack and sinatra examples already do.)

Issue #724, opened 2018-04-25
Mock out Redis in tests instead of requiring a running Redis instance


ArturMoczulski:
When I run `rake spec` on `ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]` I get the following test failure:

```
2) Rollbar::Item#dump with Redis instance in payload and ActiveSupport is enabled dumps to JSON correctly
     Failure/Error: redis.set('foo', 'bar')
     
     Redis::CannotConnectError:
       Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:344:in `rescue in establish_connection'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:329:in `establish_connection'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:99:in `block in connect'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:291:in `with_reconnect'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:98:in `connect'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:363:in `ensure_connected'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:219:in `block in process'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:304:in `logging'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis/client.rb:218:in `process'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis.rb:45:in `block in synchronize'
     # /usr/local/rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/monitor.rb:214:in `mon_synchronize'
     # /usr/local/rvm/gems/ruby-2.3.1/gems/redis-4.0.1/lib/redis.rb:45:in `synchronize'
[...]

PR #828, opened 2019-02-21
Remove dependency on redis server endpoint

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/724

It looks like the main point of this rspec example is to serialize the object, but it's this single call to `Redis#set` that requires having a live redis server.

### DIFF
--- a/lib/rollbar/plugin.rb
+++ b/lib/rollbar/plugin.rb
@@ -68,7 +68,7 @@ module Rollbar
       begin
         revert_callables.each(&:call)
       rescue StandardError => e
-        log_loading_error(e)
+        log_unloading_error(e)
       ensure
         self.loaded = false
       end
@@ -117,6 +117,10 @@ module Rollbar
 
     def log_loading_error(error)
       Rollbar.log_error("Error trying to load plugin '#{name}': #{error.class}, #{error.message}")
+    end
+
+    def log_unloading_error(error)
+      Rollbar.log_error("Error trying to unload plugin '#{name}': #{error.class}, #{error.message}")
     end
   end
 end
--- a/lib/rollbar/plugin.rb
+++ b/lib/rollbar/plugin.rb
@@ -32,7 +32,7 @@ module Rollbar
 
     def load_scoped!(transparent = false)
       if transparent
-        load! unless load?
+        load! if load?
 
         result = yield
 
--- a/lib/rollbar/plugin.rb
+++ b/lib/rollbar/plugin.rb
@@ -7,6 +7,8 @@ module Rollbar
     attr_reader :name
     attr_reader :dependencies
     attr_reader :callables
+    attr_reader :revert_callables
+    attr_accessor :on_demand
     attr_accessor :loaded
 
     private :loaded=
@@ -15,11 +17,37 @@ module Rollbar
       @name = name
       @dependencies = []
       @callables = []
+      @revert_callables = []
       @loaded = false
+      @on_demand = false
+    end
+
+    def load_on_demand
+      @on_demand = true
     end
 
     def configuration
       Rollbar.configuration
+    end
+
+    def load_scoped!(transparent = false)
+      if transparent
+        load! unless load?
+
+        result = yield
+
+        unload! if loaded
+      else
+        return unless load?
+
+        load!
+
+        result = yield
+
+        unload!
+      end
+
+      result
     end
 
     def load!
@@ -27,10 +55,22 @@ module Rollbar
 
       begin
         callables.each(&:call)
-      rescue => e
+      rescue StandardError => e
         log_loading_error(e)
       ensure
         self.loaded = true
+      end
+    end
+
+    def unload!
+      return unless loaded
+
+      begin
+        revert_callables.each(&:call)
+      rescue StandardError => e
+        log_loading_error(e)
+      ensure
+        self.loaded = false
       end
     end
 
@@ -38,8 +78,12 @@ module Rollbar
       callables << block
     end
 
-    def execute!(&block)
-      block.call if load?
+    def execute!
+      yield if load?
+    end
+
+    def revert(&block)
+      revert_callables << block
     end
 
     private
@@ -61,7 +105,7 @@ module Rollbar
 
     def load?
       !loaded && dependencies_satisfy?
-    rescue => e
+    rescue StandardError => e
       log_loading_error(e)
 
       false
@@ -71,8 +115,8 @@ module Rollbar
       dependencies.all?(&:call)
     end
 
-    def log_loading_error(e)
-      Rollbar.log_error("Error trying to load plugin '#{name}': #{e.class}, #{e.message}")
+    def log_loading_error(error)
+      Rollbar.log_error("Error trying to load plugin '#{name}': #{error.class}, #{error.message}")
     end
   end
 end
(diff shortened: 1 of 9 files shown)


jondeandres:
@waltjones the Redis dependency came from this https://github.com/rollbar/rollbar-gem/pull/254. We had problems serializing the web session when there was some TCP connection on the Rails session object. That happened for some customers depending on the storage used for the session.

The `.set()` operation I think was added in order to start a real TCP connection to the Redis server and then check the change was working correctly. So...I think removing that `.set()` operation is not safe unless we've fixed the problem in other way or we plan to change the JSON serialization library

ArturMoczulski:
Thanks for info @jondeandres . I'm going to work on this and see if it's safe to remove the actual Redis instance.

ArturMoczulski:
@jondeandres I found the root cause of the issue. Here is more info:
https://github.com/rollbar/rollbar-gem/issues/724
https://github.com/redis/redis-rb/issues/830

@waltjones this is a great improvement for the specs. However, until the Redis issue is fixed I would rather have your new test and the previous test there so it's easier for us to test if the upcoming solution for the Redis problem works as expected. Does it make sense?

waltjones:
@ArturMoczulski I agree with you on having a TCPSocket test. (Great work on tracking down the root cause by the way.) However, my testing appears to show that the original test case wasn't exercising this condition. `item.dump` _is_ actually using ActiveSupport JSON, but doesn't hang during serialization as you'd expect.

ArturMoczulski:
You mean it doesn't hang when you remove TCPSocket our of the equation? Or does it not hang for you even when serializing TCPSocket?

ArturMoczulski:
I brought up this issue on Ruby's issue tracker: https://bugs.ruby-lang.org/issues/15636

waltjones:
I have added back a test case for the Redis issue, independent of having a running redis server. The hard part (for me) was getting the original issue to repro. It turns out the specific ActiveSupport issue only occurs on ActiveSupport 4.1.x - 5.1.x. Also, it wasn't sufficient to use an instance of `Redis::Connection::TCPSocket`. For some reason, the issue only happens when the socket is wrapped by a `Redis::Connection` instance.

ArturMoczulski:
Thanks to feedback from Ruby maintainers https://bugs.ruby-lang.org/issues/15636#note-1 I understand fully what is the issue here now. `to_a` on a socket serves a different purpose then converting the instance to an array. Instead, it's supposed to read lines from the IO and split them into lines. Ruby doesn't hand in infinite recursion. It hangs waiting for input. ActiveSupport `4.1.x - 5.1.x` uses `to_a` in this particular case for the wrong reason - obviously, it's an edge case for them.

Not sure what is the best solution for this situation. I'm thinking about temporarily overriding some methods on the `::Socket` instance for the purpose of serialization and then reverting when serialization is done.

ArturMoczulski:
This is the solution I implemented:

I override `::BasicSocket#as_json` for `active_support` < `5.2` in a scoped manner ensuring that the monkey patch is only used during the serialization and doesn't creep into the user app or rest of Rollbar code: https://github.com/rollbar/rollbar-gem/issues/845

Doing this required implementing scoped plugin loading feature as described here: https://github.com/rollbar/rollbar-gem/issues/846

@waltjones @rokob please review

lib/rollbar/plugins/basic_socket.rb
```diff
@@ -1,16 +1,29 @@
...
     require 'socket'
 
-    class BasicSocket
-      def as_json(*)
-        to_s
+    class BasicSocket # :nodoc:
+      def as_json(_options = nil)
+        {
+          :value => inspect
+        }.to_json
```
waltjones: While I think the original `to_s` was fine, `#as_json` is actually expected to return a hash. So you could either drop the `.to_json`, or just return `inspect` or `to_s`. If you're going to return a string anyway, might as well make it simple.

lib/rollbar/plugins/basic_socket.rb
```diff
@@ -1,16 +1,29 @@
...
       end
     end
   end
+
+  revert do
+    ::BasicSocket.define_method(:as_json, @original_as_json)
```
waltjones: The travis builds are failing because `#define_method` is private. You'll need to use `#send` here.

lib/rollbar/plugin.rb
```diff
@@ -15,31 +17,73 @@ def initialize(name)
...
+
+    def load_on_demand
+      @on_demand = true
     end
 
     def configuration
       Rollbar.configuration
     end
 
+    def load_scoped!(transparent = false)
```
waltjones: To me, the `transparent` flag seems unnecessary, and the `transparent == false` case causes plugin dependency statements to have unintended side effects. (By causing code outside the plugin to not run.) The current use case relies on the `transparent == true` case, which makes sense. Do you have a use case in mind for the other behavior?

lib/rollbar/plugin.rb
```diff
@@ -15,31 +17,73 @@ def initialize(name)
...
       end
     end
 
+    def unload!
+      return unless loaded
+
+      begin
+        revert_callables.each(&:call)
+      rescue StandardError => e
+        log_loading_error(e)
```
waltjones: It would be nice if this had a different message when failing on unload. e.g. "Error trying to unload plugin"

lib/rollbar/plugins/basic_socket.rb
ArturMoczulski: Great catch. Thank you. I'd rather stick to what `#as_json` is supposed to return instead of returning a string.

lib/rollbar/plugins/basic_socket.rb
ArturMoczulski: Good catch. Not sure why my Ruby (`2.6.1`) didn't give me a hard time over this.

lib/rollbar/plugins/basic_socket.rb
```diff
@@ -1,16 +1,29 @@
...
   dependency { !configuration.disable_core_monkey_patch }
 
   # Needed to avoid active_support (< 4.1.0) bug serializing JSONs
-  dependency { defined?(ActiveSupport::VERSION::STRING) }
+  dependency do
+    defined?(ActiveSupport::VERSION::STRING) &&
+      Gem::Version.new(ActiveSupport::VERSION::STRING) < Gem::Version.new('5.2.0')
+  end
+
+  @original_as_json = ::BasicSocket.public_instance_method(:as_json)
```
waltjones: With the other error out of the way, this now fails for ActiveSupport <= 4.0 which doesn't define an `#as_json` method. 

The original code expected `dependency { defined?(ActiveSupport::VERSION::STRING) }` to exclude ActiveSupport < 4.1, but that doesn't seem to be working.

lib/rollbar/plugin.rb
ArturMoczulski: The `transparent == false` would be used if you want the provided logic to only be run if the plugin is successfully loaded. Which I assume would be in the majority of situations, other than this ;) Does it make sense?

lib/rollbar/plugin.rb
ArturMoczulski: 👍

lib/rollbar/plugins/basic_socket.rb
ArturMoczulski: Fixed

ArturMoczulski:
@waltjones last round of feedback and hopefully this can be merged in

lib/rollbar/plugin.rb
waltjones: Yes, thank you.

waltjones:
Yes, looks good to merge.

Issue #117, opened 2014-06-05
Gather local variables for exceptions


brianr:
Like we do in Python: https://rollbar.com/blog/post/2014/06/05/local-variable-values-in-stack-traces

albertyw:
+1

chenatuc:
+1 for this critical feature! This will save the engineers tremendous amount of time trouble-shooting issues.

waltjones:
Recently documented here: https://docs.rollbar.com/docs/ruby#section-enabling-local-variables-in-stack-traces

PR #829, opened 2019-02-23
Enable locals in exception frame data

waltjones:
https://github.com/rollbar/rollbar-gem/issues/117

This is currently a working proof of concept, to get feedback and to continue testing.

If you use this branch and enable `config.send_extra_frame_data` and `config.locals = { :enabled => true }`, local vars will be included in the frame data and will show up in the Rollbar Traceback for the exception.

The functionality relies on the Ruby `TracePoint` interface, which has been around since ~2.0, but I have found in testing so far that it works as expected in MRI 2.3.0 and up, and has best performance in 2.6.x. For now, if you want to try this in your own Rails project, use >= 2.3.0, and 2.6.1 is preferred.

### Known issues, caveats, etc.

- ~Only the Rails middleware is wrapped yet. The functionality isn't Rails specific; I just haven't wrapped the other middlewares yet.~ Rack/Sinatra middleware is enabled now too.

- A `locals` config option has been added in the style of the python option. (e.g. `config.locals = { :enabled => true }`) Currently, there is just the enable option. It's expected that more of the additional config options used in python, or similar, should be added at some point.

- The values of local vars are added to the payload using `#to_s`, to avoid ActiveSupport serialization issues. More can be done here if we decide to move forward with the feature. Similar to the python feature, objects could be fully or partially serialized or just identified by class name.

- ~No scrubbing is implemented yet.~ The params scrubber is now applied to locals.

### Performance

TracePoint is a thin wrapper around `Kernel#set_trace_func` and is more performant than `set_trace_func` largely because you can avoid hooking all events unnecessarily. But also, TracePoint has become more performant in later versions of Ruby, especially 2.6.x. 

Test timings were generated using a test code path that generated over 20K method calls, 40K block entry calls, and 20K C function calls.

The rake task to generate this data is here: 
https://github.com/rollbar/rollbar-gem/blob/wj-exception-locals/lib/tasks/benchmark.rake

Benchmarks for the instrumented code path are here:
https://docs.google.com/spreadsheets/d/1keJNwA2vHvyCZHKilvcB2JnPRzlvoiZBUHy9lK2ppkk/edit#gid=429472357

And summarized:

Ruby | Without hooks | With hooks | Full locals capture | Production (test branching removed)
-----|---------------|------------|---------------------|-----------
2.3.0 | 0.0100s | 0.0300s | 0.1200s |
2.6.1 | 0.0048s | 0.0292s | 0.1040s | 0.0745s
 
Concerns about peformance can be mitigated in several ways. Currently the hook is only enabled after the exception middleware in the request path. It could also be enabled only for specific requests. The design of the integration to rollbar-gem can allow on-the-fly enable/disable, which could benefit specific production debugging scenarios. Finally, in Ruby 2.6.x, TracePoint has granular enable filters (selecting a single method, for example) that are meant to preserve performance, even in production.

It's not clear whether or not this feature is more or less performant than the python implementation, or by how much. User expectations for that environment may be able to help inform decisions here.

### DIFF
--- a/lib/rollbar/item/locals.rb
+++ b/lib/rollbar/item/locals.rb
@@ -1,4 +1,5 @@
 require 'rollbar/notifier'
+require 'rollbar/scrubbers/params'
 
 module Rollbar
   class Item
@@ -9,10 +10,8 @@ module Rollbar
         end
 
         def locals_for_location(filename, lineno)
-          puts "Location: #{filename}:#{lineno}"
           if (frame = frame_for_location(filename, lineno))
-            puts "Frame: #{frame[:path]}:#{frame[:lineno]}"
-            locals_for(frame[:binding]).tap { |locals| puts locals.inspect }
+            scrub(locals_for(frame[:binding]))
           else
             {}
           end
@@ -22,8 +21,6 @@ module Rollbar
           while (frame = exception_frames.pop)
             return nil unless frame
             return frame if matching_frame?(frame, filename, lineno)
-
-            puts "Skipped Frame: #{frame[:path]}:#{frame[:lineno]}"
           end
           nil
         end
@@ -43,7 +40,15 @@ module Rollbar
         end
 
         def prepare_value(value)
-          value.inspect
+          value.to_s
+        end
+
+        def scrub(hash)
+          Rollbar::Scrubbers::Params.call(
+            :params => hash,
+            :config => Rollbar.configuration.scrub_fields,
+            :whitelist => Rollbar.configuration.scrub_whitelist
+          )
         end
       end
     end
--- a/lib/rollbar/item/locals.rb
+++ b/lib/rollbar/item/locals.rb
@@ -28,6 +28,8 @@ module Rollbar
           nil
         end
 
+        private
+
         def matching_frame?(frame, filename, lineno)
           frame[:path] == filename && frame[:lineno].to_i <= lineno.to_i
         end
--- a/lib/rollbar/item/locals.rb
+++ b/lib/rollbar/item/locals.rb
@@ -1,0 +1,48 @@
+require 'rollbar/notifier'
+
+module Rollbar
+  class Item
+    class Locals # :nodoc:
+      class << self
+        def exception_frames
+          Rollbar.notifier.exception_bindings
+        end
+
+        def locals_for_location(filename, lineno)
+          puts "Location: #{filename}:#{lineno}"
+          if (frame = frame_for_location(filename, lineno))
+            puts "Frame: #{frame[:path]}:#{frame[:lineno]}"
+            locals_for(frame[:binding]).tap {|locals| puts locals.inspect }
+          else
+            {}
+          end
+        end
+
+        def frame_for_location(filename, lineno)
+          while (frame = exception_frames.pop)
+            return nil unless frame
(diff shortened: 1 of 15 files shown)


rokob:
@jondeandres This is the work for local variables

lib/rollbar/item/locals.rb
```diff
@@ -0,0 +1,49 @@
...
+          puts "Location: #{filename}:#{lineno}"
+          if (frame = frame_for_location(filename, lineno))
+            puts "Frame: #{frame[:path]}:#{frame[:lineno]}"
+            locals_for(frame[:binding]).tap { |locals| puts locals.inspect }
+          else
+            {}
+          end
+        end
+
+        def frame_for_location(filename, lineno)
```
ArturMoczulski: Consider hiding some of those methods behind `protected` or `private` since they are only used internally. What do you think?

lib/rollbar/middleware/rails/rollbar.rb
```diff
@@ -21,6 +21,7 @@ def call(env)
 
           Rollbar.scoped(scope) do
             begin
+              Rollbar.notifier.enable_locals
```
ArturMoczulski: What's the reason behind enabling and disabling here? Is it that you want this to be the default behavior only for Rails?

ArturMoczulski:
Looks great so far. Just a couple of minor questions in the review. Let me know when you want me to do the full approval :)

lib/rollbar/item/locals.rb
waltjones: Great idea. I'll do that.

lib/rollbar/middleware/rails/rollbar.rb
waltjones: These will be added to the other middleware as well. The intent is to only have the TracePoint hooks turned on when needed, to minimize the perf impact.

ArturMoczulski: Oh yeah you mentioned that. Awesome

waltjones:
@ArturMoczulski thank you! I'm still considering this a WIP for now, and getting feedback.

ArturMoczulski:
What is there left to do? Testing other use cases than Rails?

waltjones:
Some kind of scrub policy should be added before merge. I'd also like to improve the serialization options, but perhaps that can go in a later PR. This all would be a good topic for SDK mtg today.

rokob:
Overall I really like this approach. TracePoint was designed for this and it is already part of the bytecode so we are not hacking too much. We can dig in to the implementation details more, but for now I'd say this is great to have that people can turn on without too crazy of a perf impact. I'm not convinced that we can get to a place where this defaults to on for everyone from a performance point of view, but maybe. I wouldn't mind writing part of this in C if we could actually benefit and not get swamped by the bridge cost.

The Ruby version will never be as "performant" as the Python version because this work of keeping the stack of locals is done implicitly by both VMs, but Python gives you a function to turn the current local table into a dict. In Ruby if we could get a handle to the current `iseq` and call `rb_iseq_local_variables` from `vm.c`, and turn that result into a hash, that would be roughly what Python is doing with `locals()`. But that isn't portable.

Let's push this approach forward and we can work out the implementation details.

Issue #832, opened 2019-02-25
Capistrano deploy fails if No/incorrect Access token is provided in version 2.19.x


rdunlop:
I upgraded to 2.19.1, and my `cap production deploy` started failing. This was working on 2.18.2.

Investigating this, I found that I was not specifying a `rollbar_token` properly in the `deploy.rb` file.

Yes, I agree that finding this out was good for my project, though a warning may have been sufficient.

BUT, I feel that this is a breaking change, and should not have been introduced without more warning in the changelog (and perhaps a major version bump). I suspect that some people may want to be able to perform deploys to some environments without informing rollbar...so perhaps this wasn't introduced on purpose.

Summary:
- If you don't have a `rollbar_token` specified properly, capistrano deploy will fail
- If you do have a `rollbar_token` specified, but it is invalid, capistrano deploy will fail.

Error messages:

Without token:
```
00:10 rollbar:deploy_started
      Notifying Rollbar of deployment start
      #<URI::HTTPS https://api.rollbar.com/api/1/deploy/>: {"access_token":null,"environment":"production","revision":"90a4de294fcc1953d0a6f5371abc6ef30c705485","rollbar_username":null,"local_username":null,"comment":null,"status":"started","proxy":"ENV","dry_run":false}
      400; Bad Request; {  "err": 1,  "message": "access token required"}
#<Thread:0x000055658df84768@/home/circleci/tmp/vendor/bundle/gems/sshkit-1.18.0/lib/sshkit/runners/parallel.rb:10 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	1: from /home/circleci/tmp/vendor/bundle/gems/sshkit-1.18.0/lib/sshkit/runners/parallel.rb:11:in `block (2 levels) in execute'
/home/circleci/tmp/vendor/bundle/gems/sshkit-1.18.0/lib/sshkit/runners/parallel.rb:15:in `rescue in block (2 levels) in execute': Exception while executing as <redacted>: undefined method `[]' for nil:NilClass (SSHKit::Runner::ExecuteError)
(Backtrace restricted to imported tasks)
cap aborted!
SSHKit::Runner::ExecuteError: Exception while executing as <redacted>: undefined method `[]' for nil:NilClass


Caused by:
NoMethodError: undefined method `[]' for nil:NilClass

Tasks: TOP => rollbar:deploy_started
(See full trace by running task with --trace)
[...]

PR #837, opened 2019-02-26
Issue #832: Capistrano hotfix

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/832

Hopefully, this will hotfix the current Capistrano issues.

The Capistrano integration code needs RSpec coverage soon.

### DIFF
--- a/lib/rollbar/capistrano3.rb
+++ b/lib/rollbar/capistrano3.rb
@@ -10,21 +10,21 @@ namespace :rollbar do
   desc 'Send deployment started notification to Rollbar.'
   task :deploy_started do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_started(self, self, dry_run?)
+      ::Rollbar::CapistranoTasks.deploy_started(self, self, false)
     end
   end
 
   desc 'Send deployment succeeded notification to Rollbar.'
   task :deploy_succeeded do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_succeeded(self, self, dry_run?)
+      ::Rollbar::CapistranoTasks.deploy_succeeded(self, self, false)
     end
   end
 
   desc 'Send deployment failed notification to Rollbar.'
   task :deploy_failed do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_failed(self, self, dry_run?)
+      ::Rollbar::CapistranoTasks.deploy_failed(self, self, false)
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/rollbar/capistrano3.rb
```diff
@@ -10,21 +10,21 @@
   desc 'Send deployment started notification to Rollbar.'
   task :deploy_started do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_started(self, self, dry_run?)
+      ::Rollbar::CapistranoTasks.deploy_started(self, self, false)
```
waltjones: I didn't see/couldn't find a valid `dry_run` option (though I'm sure one exists), so have set to `false` for now.

lib/rollbar/capistrano_tasks.rb
```diff
@@ -15,7 +15,7 @@ def deploy_started(capistrano, logger, dry_run)
           capistrano.set(:rollbar_deploy_id, 123) if dry_run
 
           skip_in_dry_run(logger, dry_run) do
-            if (deploy_id = result[:data][:deploy_id])
+            if (deploy_id = result[:data] && result[:data][:deploy_id])
```
waltjones: This originally failed because `result` was null, but it can't hurt to guard `result[:data]` as well.

lib/rollbar/capistrano_tasks.rb
```diff
@@ -123,6 +123,10 @@ def skip_in_dry_run(logger, dry_run)
       end
 
       def info_request_response(logger, result)
+        # TODO: Is this logger output wanted or expected?
+        # Should there be a switch to disable?
+        return
```
waltjones: See related github issue for discussion about this logger output. It looks like it includes output that needs to be scrubbed, and maybe should be debug level when turned on.

lib/rollbar/capistrano3.rb
ArturMoczulski: What's the reason behind changing this? `dry_run?` is a method from the Capistrano3 object which should be available in the context of self here. It's set in the command line when running capistrano `--dry-run`.

ArturMoczulski: Oh, I see: https://github.com/rollbar/rollbar-gem/issues/832#issuecomment-467522316

That's interesting. I'm investigating this.

ArturMoczulski:
Merged and will be released in `v2.19.2`. However, the fix needs some work to correctly implemented support for `--dry-run`. The continuation of the work will be performed under the following issue: https://github.com/rollbar/rollbar-gem/issues/839

PR #835, opened 2019-02-25
Add rubygems links

viraptor:
Add direct links to the gem metadata.
This allows better automatic dependency management.

PR #838, opened 2019-02-26
link to source code from gemspec

grosser:
useful for automated tools that try to show gem diff / the new way of setting a source-code url now that rubygems.org has removed these fields


waltjones:
@grosser thank you for the PR, but we already have this open PR that includes the same change: https://github.com/rollbar/rollbar-gem/pull/835

Issue #839, opened 2019-02-27
Capistrano deploy notification fails due to missing dry_run? method in some cases


ArturMoczulski:
This issue was brought up in the course of working on https://github.com/rollbar/rollbar-gem/issues/832

For now the hotfix just forces `dry_run` to `false`, but the correct propagation of the `dry_run?` option still needs to be implemented.

In cases of user @innocode-devops an error is thrown:

```
* Invoke rollbar:deploy_started (first_time)
** Execute rollbar:deploy_started
cap aborted!
SSHKit::Runner::ExecuteError: Exception while executing as dev@host.com: undefined method `dry_run?' for #<SSHKit::Backend::Netssh:0x000055a19ad675e0>
/usr/local/bundle/gems/sshkit-1.18.2/lib/sshkit/runners/parallel.rb:15:in `rescue in block (2 levels) in execute'
/usr/local/bundle/gems/sshkit-1.18.2/lib/sshkit/runners/parallel.rb:11:in `block (2 levels) in execute'

Caused by:
NoMethodError: undefined method `dry_run?' for #<SSHKit::Backend::Netssh:0x000055a19ad675e0>
/usr/local/bundle/gems/rollbar-2.19.0/lib/rollbar/capistrano3.rb:13:in `block (3 levels) in <top (required)>'
/usr/local/bundle/gems/sshkit-1.18.2/lib/sshkit/backends/abstract.rb:29:in `instance_exec'
/usr/local/bundle/gems/sshkit-1.18.2/lib/sshkit/backends/abstract.rb:29:in `run'
/usr/local/bundle/gems/sshkit-1.18.2/lib/sshkit/runners/parallel.rb:12:in `block (2 levels) in execute'
Tasks: TOP => rollbar:deploy_started
The deploy has failed with an error: Exception while executing as dev@host.com: undefined method `dry_run?' for #<SSHKit::Backend::Netssh:0x000055a19ad675e0>
** Invoke deploy:failed (first_time)
** Execute deploy:failed
** Invoke rollbar:deploy_failed (first_time)
** Execute rollbar:deploy_failed
```

This is unusual, as such issue has not been brought up during testing and I'm still unable to reproduce this. From what I understand the `dry_run?` method should be available in the Capistrano object coming from `Capistrano::DSL::Env` through the `Forwardable` pattern. That is what happens during my local testing. However, in case of @innocode-devops the method seems to be missing.

ArturMoczulski:
[...]

PR #840, opened 2019-02-27
Issue #839: Capistrano deploy notification fails due to missing dry_run? method in some cases

ArturMoczulski:
Solution for https://github.com/rollbar/rollbar-gem/issues/839

The `dry_run?` helper was not introduced until Capistrano 3.5. For backward Capistrano compatibility, this PR uses the old style of determining the `--dry-run` mode which is also still compatible with the latest Capistrano releases

### DIFF
--- a/lib/rollbar/capistrano3.rb
+++ b/lib/rollbar/capistrano3.rb
@@ -7,24 +7,27 @@ require 'rollbar/deploy'
 require 'rollbar/capistrano_tasks'
 
 namespace :rollbar do
+  # dry_run? wasn't introduced till Capistrano 3.5.0; use the old fetch(:sshkit_backed)
+  set :dry_run, (proc { ::Capistrano::Configuration.env.fetch(:sshkit_backend) == ::SSHKit::Backend::Printer })
+
   desc 'Send deployment started notification to Rollbar.'
   task :deploy_started do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_started(self, self, false)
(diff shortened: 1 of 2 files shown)


lib/rollbar/capistrano3.rb
```diff
@@ -7,24 +7,27 @@
...
 
 namespace :rollbar do
+  # dry_run? wasn't introduced till Capistrano 3.5.0; use the old fetch(:sshkit_backed)
+  set :dry_run, (proc { ::Capistrano::Configuration.env.fetch(:sshkit_backend) == ::SSHKit::Backend::Printer })
+
   desc 'Send deployment started notification to Rollbar.'
   task :deploy_started do
     on primary fetch(:rollbar_role) do
-      ::Rollbar::CapistranoTasks.deploy_started(self, self, false)
+      ::Rollbar::CapistranoTasks.deploy_started(self, self, fetch(:dry_run))
```
waltjones: The previous call had been `dry_run?`, and here you are not using the `?`. Maybe that is why the other one was failing? Otherwise, does this need to have the `?` as well?

ArturMoczulski: No. Previously this was failing for a user on Capistrano 3.4 which didn't have the `dry_run?` helper method. It was introduced in Capistrano 3.5. Because of this I need to manually verify the dry run using the same logic that was used before 3.5 and set it in `dry_run` property which is what I'm doing on line 11. Then I `fetch` that property.

Does it make sense?

waltjones: Yes, and I overlooked line 11. Thank you for the explanation!

Issue #620, opened 2017-08-07
Gem failsafe error about connecting to DB


jessewgibbs:
Details in https://app.clubhouse.io/rollbar/story/22078/gem-failsafe-error-about-connecting-to-db

rivkahstandig3636:
Closing this issue as we are unable to reproduce it.

sdhull:
We've actually been seeing this intermittently in production as well. The primary issue is that the error has basically no information. I have no idea how to go about debugging this. What's a guy gotta do to get backtrace?

`Failsafe from rollbar-gem. ActiveRecord::ConnectionTimeoutError: \"could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use\" in /app/vendor/bundle/ruby/2.5.0/gems/activerecord-5.2.2.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:201:in 'block in wait_poll': build_item in exception_data'`

brianr:
Thanks @sdhull .

@waltjones Any ideas on 1) what might cause this situation, and 2) how rollbar-gem can better handle it?

waltjones:
@sdhull I'll take a look to see if more diagnostic info can be provided in these failsafe conditions.

@brianr It looks like the original error is the connection pool is exhausted, with unknown root cause. It looks like rollbar-gem should provide more info on failsafe conditions regardless of what error got us there.

brianr:
@waltjones 👍 

The other thing I'm wondering is whether this situation can result in a more positive outcome than a failsafe. I assume the db connection attempt is happening when rollbar-gem is trying to get the person data, or maybe in "custom_data_fn" or something like that. If that happens, I'd think we can still report the original stack trace and the contextual data that we are able to get; we just won't have the person data/whatever failed to build.

waltjones:
@brianr Good idea.

sdhull:
Super impressed how responsive you guys are. Thanks! Looking forward to seeing this move

rokob:
Usually by the time we get to a failsafe it is because we tried making a payload with some information and even that failed. If there is an exception in say the person function we should already just ignore it and try to move on with that blank.

waltjones:
@sdhull can you bundle with branch `wj-failsafe`? This isn't a final fix, but should allow you to get the report for the original exception. If the original non-rollbar exception is also `ActiveRecord::ConnectionTimeoutError` that would be interesting to know.

PR #849, opened 2019-03-15
Tolerate connection pool errors

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/620

If ActiveRecord times out while getting person data, rescue the exception and return an empty person object. This allows sending the original exception report, rather than an internal error report.


sdhull:
It might be nice to add information to the exception report that the ActiveRecord proc to get the person data timed out.

While this is probably better that present-day (in that it sends the original exception report), it does lose some information (namely -- you have some problem with AR connection pool). I'm not sure how exactly you might add that to the exception report from this context, but imo that would be ideal

waltjones:
I agree.  Maybe it makes sense to generate two reports. As far as I know, there's not a way to send two exceptions in one report.

Issue #850, opened 2019-03-17
Some of the `basic_socket` specs fail randomly and temporarily


ArturMoczulski:
Some of the time the spec `basic_socket plugin #load_scoped! with core monkey patching enabled using active_support < 5.2 changes implementation of ::BasicSocket#as_json temporarily` fails in Travis builds. Usually, restarting the job fixes the problem, but nevertheless, the jobs sometimes run into this problem:

When comparing the pre-`load_scoped!` `::BasicSocket#as_json`'s source location with the post-`load_scoped!` source location, the two sometimes don't match and the test fails.

PR #851, opened 2019-03-18
Use alias_method to replace #as_json

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/850

`BasicSocket#define_method` causes "TypeError: bind argument must be an instance of Enumerable".

I'm seeing this in JRuby specifically, and when the `#load_scoped!` spec runs in isolation or runs before the `#load!` spec. So, using JRuby (I'm using 9.1.9.0), either of these will repro 100%:

bundle exec rspec spec/rollbar/plugins/basic_socket_spec.rb:35

bundle exec rspec --seed 26852 spec/rollbar/plugins/basic_socket_spec.rb

It turns out `#unload!` always fails in JRuby, but running the specs in a different order will mask the failure.

Root cause is:
* ActiveSupport adds `#as_json` to `Enumerable`, which is included in `BasicSocket`.
* The original UnboundMethod that is saved to be bound later has `#owner == Enumerable` on it.
* `BasicSocket#define_method` fails with "TypeError: bind argument must be an instance of Enumerable" because JRuby `#define_method` can't see that BasicSocket is an instance of enumerable.

On any Ruby, unbound methods can only be bound to the same type as their original owner. In JRuby, the binding won't work because the owner is `Enumerable`.

I tried a variety of things, none of which would work in JRuby.
```
@original_as_json.bind(@original_as_json.owner)
```
uses the actual owner from the `UnboundMethod` object, but fails with the same exception.

And, here I tried converting the `UnboundMethod` to a proc (which is said to work in MRI), but it fails with "TypeError: no implicit conversion of UnboundMethod into Proc".
```
::BasicSocket.send(:define_method, :as_json, &@original_as_json)
```

The final solution using `alias_method` should work on all Rubies, and doesn't need to know how `#as_json` was originally added or by what module.

### DIFF
--- a/spec/rollbar/plugins/basic_socket_spec.rb
+++ b/spec/rollbar/plugins/basic_socket_spec.rb
@@ -14,6 +14,10 @@ end
 describe 'basic_socket plugin' do
   subject { Rollbar.plugins.get('basic_socket') }
 
+  before(:all) do
+    Rollbar.plugins.get('basic_socket').unload!
+  end
+
   after(:each) do
     subject.unload!
   end
@@ -38,11 +42,13 @@ describe 'basic_socket plugin' do
               source_location
 
             subject.load_scoped! do
+              expect(subject.loaded).to eq(true)
               socket = TCPSocket.new 'example.com', 80
               expect(socket.as_json).to include(:value)
               expect(socket.as_json[:value]).to match(/TCPSocket/)
             end
 
+            expect(subject.loaded).to eq(false)
             expect(BasicSocket.public_instance_method(:as_json).source_location).
               to(eq(original_implementation))
           end
@@ -54,8 +60,10 @@ describe 'basic_socket plugin' do
               result = false
               subject.load_scoped!(true) do
                 result = true
+                expect(subject.loaded).to eq(false) # Plugin should not load
               end
               expect(result).to eq(true)
+              expect(subject.loaded).to eq(false)
             end
           end
         end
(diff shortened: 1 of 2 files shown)


waltjones:
Specs updated to ensure the plugin starts out unloaded. If the JSON `Item#dump` spec `it 'fails in ActiveSupport with stack too deep'` runs before these specs, it will leave the plugin in a loaded state. If then the "doesn't load by default" spec runs first, it will fail since the plugin is still loaded.

While the plugin code is written to ensure unload, the JSON item dump spec intentionally causes one of `NoMemoryError, SystemStackError, Java::JavaLang::StackOverflowError` and I'm assuming because they are system errors, the cleanup code doesn't run. 

I also added some expectations for unload, and aim to consistently unload (one way or another) at the end of each example.

spec/rollbar/plugins/basic_socket_spec.rb
```diff
@@ -14,6 +14,10 @@
 describe 'basic_socket plugin' do
   subject { Rollbar.plugins.get('basic_socket') }
 
+  before(:all) do
+    Rollbar.plugins.get('basic_socket').unload!
+  end
+
   after(:each) do
     subject.unload!
   end
```
waltjones: I'd be OK removing this `after` block, as each example now performs unload. I'm also OK keeping it.

spec/rollbar/plugins/basic_socket_spec.rb
```diff
@@ -14,6 +14,10 @@
 describe 'basic_socket plugin' do
   subject { Rollbar.plugins.get('basic_socket') }
 
+  before(:all) do
+    Rollbar.plugins.get('basic_socket').unload!
+  end
```
waltjones: `before(:all)` can't use `subject`, so the full thing is expanded here.

ArturMoczulski:
LGTM. Feel free to merge

PR #854, opened 2019-03-19
Enable travis builds to test rails runner process

waltjones:
This spec had been marked pending because `ENV['BUNDLE_GEMFILE']` doesn't work as expected in Travis. The workaround here is to set a different env var in the gemfile. This works, and the complete matrix coverage exposed a couple issues that are also fixed here:

- [x] Rails 5.0.x needs to use the old runner, only 5.1.x and up should use the new Rails 5 runner.
- [x] JRuby doesn't have `fork` and can't use the rspec-command gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,13 @@
 
 source 'https://rubygems.org'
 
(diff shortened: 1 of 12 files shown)


ArturMoczulski:
LGTM

PR #862, opened 2019-04-23
Add Conventional Commits to Contributor section of README

waltjones:
Fixes https://github.com/rollbar/rollbar-gem/issues/860

This does _not_ add commitizen to project dependencies, based on nodejs.org recommendation:

> If you’re installing something that you want to use in your program, using require('whatever'), then install it locally, at the root of your project.

> If you’re installing something that you want to use in your shell, on the command line or something, install it globally, so that its binaries end up in your PATH environment variable.

https://nodejs.org/en/blog/npm/npm-1-0-global-vs-local-installation/

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,32 @@ For bug reports, please [open an issue on GitHub](https://github.com/rollbar/rol
 
 1. Fork it
 2. Create your feature branch (```git checkout -b my-new-feature```).
-3. Commit your changes (```git commit -am 'Added some feature'```)
+3. Commit your changes (See Conventional Commits below)
 4. Push to the branch (```git push origin my-new-feature```)
 5. Create new Pull Request
 
 We're using RSpec for testing. Run the test suite with ```rake spec```. Tests for pull requests are appreciated but not required. (If you don't include a test, we'll write one before merging.)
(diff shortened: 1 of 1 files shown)


ArturMoczulski:
In `rollbar-php` I added commitizen as a dev dependency. Do you think that's not a good idea?

waltjones:
I think it has to added globally to use it, and I also needed a `~/.czrc` for things to work. So I don't know if there's any advantage to also adding as a dev dependency.

ArturMoczulski:
Cool. Let me double check how I did it for PHP

ArturMoczulski:
Ok, I agree with. I will use your README.md guidelines to be used in the PHP SDK as well

ArturMoczulski:
Also I would an example in the contrib section how to structure the commit message if there are breaking changes. This is all in the conventional changelog docs, but I'd still add it. What do you think?

ArturMoczulski:
I made those modifications in the PHP repo: https://github.com/rollbar/rollbar-php/pull/450

brianr:
Closing; Conventional Commits hasn't been adopted in this repo at this point.

Issue #813, opened 2019-01-30
Rollbar should offer option to write payload to application log


sdhull:
Honestly I cannot for the life of me understand why #425 was merged.

There were 2 issues in the past asking for _opposite_ default behavior (#63 & #66 respectively).

A discussion should have ensued to figure out what offers the most flexibility and what is the desired default behavior. (Instead, the first group of people were told "don't worry, this is already how it works" and the second group of people were told "don't worry, we'll change how it works.")

When the default behavior was to log even when disabled, it's trivial for a developer to both set `config.enabled = false; config.default_logger = Logger.new(IO::NULL)`.
However, now that the default behavior is to suppress logging when Rollbar is disabled, it's actually impossible to get out of this behavior.

If people agree that disabling Rollbar should also set the logger to a null logger to suppress logs from rollbar, I think it's reasonable to implement that, but in a way that would allow those of us _do_ want logging still to bring it back. Maybe add `config.disable!` that also sets the default logger to `IO::NULL`.

But as it stands, this is essentially broken. I want Rollbar reporting off in development, but obviously I'm sending **important debugging info** to Rollbar and obviously _I need important debugging info_ in development. IMO the _correct_ default behavior is to log even when disabled. If people want to disable logging, there's a reasonable way to do it (set the logger to `IO::NULL`).

If maintainers agree, I'd be happy to submit a PR that unwinds #425.

sdhull:
Ugh since we have Rollbar disabled, I wasn't even clear on _what kind of logging_ it does. After enabling it locally I realize this still isn't really what I want. I don't care about metadata about the stuff I send to Rollbar, I care about the actual content that was sent.

Basically, if I have some code

```ruby
begin
  do_something_that_might_raise
rescue => e
  Rollbar.error(e)
end
```
And I have Rollbar disabled in development, I will never know if my method raised. This hampers development & is at odds with having a single way to track errors. My expectation of this tool is that it will not get in the way of being a productive developer. Obviously I could add `raise e if Rails.env.development?` everywhere I use `Rollbar.error` but that's at odds with DRY practices & a stupid thing to have to explain to new developers.

So I guess it's going to take more discussion about good default behavior and more work to implement, since it's not exactly just about unwinding #425

jessewgibbs:
@sdhull thanks for the detailed summary of the issue.  #425 pre-dates the current maintainer team, but the folks involved are still here so we can definitely pull them into a discussion.

[...]

Issue #708, opened 2018-04-02
option to re-raise errors in some environments


jjb:
I would like the option to have rollbar raise errors that are passed in to `Rollbar.error` when in the dev environment. Would the project be amenable to such a feature?

(I thought I proposed this before and we had some discussion but I coudn't find it anywhere, it may have been in a support ticket. If you feel like taking a quick look for a support ticket under my name from many years ago, that might be beneficial).

PR #867, opened 2019-05-22
Dev config flags: raise_on_error, log_payload, transmit, logger_level

waltjones:
Implements the dev flags recently defined for all SDKs.

The `verbose` flag has already been implemented in this SDK as `logger_level`, and this remains unchanged. Further, the existing default `logger_level` setting of `:info` is retained. Finally, in keeping with the expected way of disabling config settings, setting a falsey value for `logger_level` disables the verbose logging completely.

Fixes https://github.com/rollbar/rollbar-gem/issues/708
Fixes https://github.com/rollbar/rollbar-gem/issues/813

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -63,6 +63,7 @@ module Rollbar
     attr_reader :send_extra_frame_data
     attr_accessor :use_exception_level_filters_default
     attr_accessor :proxy
+    attr_accessor :raise_on_error
 
     attr_reader :project_gem_paths
 
@@ -130,6 +131,7 @@ module Rollbar
       @project_gem_paths = []
       @use_exception_level_filters_default = false
       @proxy = nil
+      @raise_on_error = false
       @collect_user_ip = true
       @anonymize_user_ip = false
       @hooks = {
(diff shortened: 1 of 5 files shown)


ArturMoczulski:
@waltjones what do you think about adding `verbose` as an alias for `logger_level` to remain the standard across SDKs?

tisba:
I found this PR by accident.

I was wondering why the config generator does not mention these options, nor does the official documentation at https://docs.rollbar.com/docs/ruby.

I'm trying to figure out, how I can log everything in dev to the application log in development, without sending data to Rollbar. Can someone give me a hint? Thanks!

waltjones:
@tisba Yes, the docs and the generator are out of date. I've opened https://github.com/rollbar/rollbar-gem/issues/948 to track this.

I think the config options you want are:
```
config.enabled = true
config.transmit = false
config.log_payload = true
config.logger = your_logger # e.g. Rails.logger
config.logger_level = level # :debug, :info, :warn, or :error
```

tisba:
That does not seem to work.

If I make these adjustments (using `logger_level = :debug`, I get this:

```
Rollbar.error("this is a test", foo: "bar")
```

returns `"disabled"` and nothing gets logged.

ponyesteves:
Seems you an `access_token` is always required .. even for logging

https://github.com/rollbar/rollbar-gem/blob/c1a2a00e8654e561f265ad71abc1732cd910ea3a/lib/rollbar/notifier.rb#L203

PR #868, opened 2019-05-23
Properly test request for JSON mime types

### LINKED ISSUE #746: JSON body params not available in rollbar
No idea what it's related to exactly and I don't have time to debug the gem unfortunately, however I can tell you this:
* we're using ember.js on the frontend
* `ACCEPT` & `CONTENT_TYPE` headers are both set to "application/vnd.api+json"
* `params` shown in rollbar only reflect route information (controller, action, id) and are completely missing the JSON from the request body
* rollbar gem version 2.15.5

scudco:
* Test `Content-Type` more liberally
* Use the correct value for `Accept` within the rack request

Fixes #746

### DIFF
--- a/spec/rollbar/request_data_extractor_spec.rb
+++ b/spec/rollbar/request_data_extractor_spec.rb
@@ -238,7 +238,19 @@ describe Rollbar::RequestDataExtractor do
                                   :method => 'POST')
       end
 
-      it 'extracts the correct user IP' do
+      it 'extracts the correct body' do
+        result = subject.extract_request_data_from_rack(env)
+        expect(result[:body]).to be_eql(body)
+      end
+
+      it 'extracts the correct body for JSONAPI' do
+        env['CONTENT_TYPE'] = 'application/vnd.api+json'
+        result = subject.extract_request_data_from_rack(env)
+        expect(result[:body]).to be_eql(body)
+      end
+
+      it 'extracts the correct body for any JSON compatible MIME type' do
+        env['CONTENT_TYPE'] = 'application/ld+json'
         result = subject.extract_request_data_from_rack(env)
         expect(result[:body]).to be_eql(body)
       end
@@ -254,7 +266,7 @@ describe Rollbar::RequestDataExtractor do
                                   :method => 'DELETE')
       end
 
-      it 'extracts the correct user IP' do
+      it 'extracts the correct body' do
         result = subject.extract_request_data_from_rack(env)
         expect(result[:body]).to be_eql(body)
       end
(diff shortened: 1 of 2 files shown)


scudco:
I'd be happy to write a test for this, but I couldn't immediately get the tests running. Is there a trick beyond `bundle && rake spec`?

sdhull:
It's worth noting that problems were caused by the mimetype for [jsonapi](https://jsonapi.org/#mime-types) which was not being recognized by `json_request?` as json

waltjones:
@scudco The root Gemfile does work, but its rails is set to 4.2.8, which doesn't work with every Ruby. The Gemfile is set up so you should be able to change to a known good rails version for your Ruby and `rm Gemfile.lock && bundle install` with no other changes.

If you're seeing a different kind of error, let me know.

scudco:
thanks @waltjones 

I tracked it down to how the support files are loaded. 

https://github.com/rollbar/rollbar-gem/blob/master/spec/spec_helper.rb#L51

`Dir.glob/[]` doesn't guarantee the order of the array afaik. Some of the files rely on classes defined within other files. Specifically, in my case, `spec/support/deploy_api/report.rb` ends up being `require`d before `spec/support/rollbar_api.rb`, but `spec/support/deploy_api/report.rb` depends on `RollbarAPI`.

```
NameError:
  uninitialized constant RollbarAPI
# ./spec/support/deploy_api/report.rb:4:in `<module:DeployAPI>'
# ./spec/support/deploy_api/report.rb:3:in `<top (required)>'
# ./spec/spec_helper.rb:52:in `block in <top (required)>'
# ./spec/spec_helper.rb:52:in `each'
# ./spec/spec_helper.rb:52:in `<top (required)>'
# ./spec/rollbar_spec.rb:13:in `<top (required)>'
```

I can work my way around this to write tests for this PR and open another issue for this specs-specific problem.

waltjones:
@scudco Thanks for the detail!

scudco:
I wrote some tests @waltjones

I wasn't 100% sure where to put them. Let me know if they belong in some other part of the test suite.

spec/rollbar/request_data_extractor_spec.rb
```diff
@@ -270,7 +282,7 @@ class ExtractorDummy
                                   :method => 'PUT')
       end
 
-      it 'extracts the correct user IP' do
+      it 'extracts the correct body' do
```
sdhull: nice

waltjones:
@scudco Tests look good. Perfect spot for them, I'd say. Thank you for the PR.

waltjones:
@scudco Released in 2.20.0

PR #871, opened 2019-05-28
Ensure RollbarAPI is loaded prior to being referenced

### LINKED ISSUE #869: Specs may not always run out-of-the-box
As first documented here: https://github.com/rollbar/rollbar-gem/pull/868#issuecomment-495745241

Specs would not run for me out-of-the-box because for whatever reason _my_ machine did not produce the same result from `Dir.[]` as others' machines. `Dir.glob/[]` doesn't guarantee the order of the returned array afaik.

Some of the files rely on classes defined within other files. Specifically, in my case, `spec/support/deploy_api/report.rb` ends up being `require`d before `spec/support/rollbar_api.rb`, but `spec/support/deploy_api/report.rb` depends on `RollbarAPI`.

```
NameError:
  uninitialized constant RollbarAPI
# ./spec/support/deploy_api/report.rb:4:in `<module:DeployAPI>'
# ./spec/support/deploy_api/report.rb:3:in `<top (required)>'
# ./spec/spec_helper.rb:52:in `block in <top (required)>'
# ./spec/spec_helper.rb:52:in `each'
# ./spec/spec_helper.rb:52:in `<top (required)>'
# ./spec/rollbar_spec.rb:13:in `<top (required)>'
```

It seems there could be different ways of addressing this problem but I'm not sure what the gem maintainers would prefer.

1. rename files to have an order and sort the glob results
2. explicitly require files that may be dependencies for files within `spec/support` then perform the glob require
3. something else better?

scudco:
Fixes #869

### DIFF
--- a/spec/support/deploy_api/report.rb
+++ b/spec/support/deploy_api/report.rb
@@ -1,4 +1,5 @@
 require 'rack/request'
+require "#{File.dirname(__FILE__)}/../rollbar_api.rb"
 
 module DeployAPI
   class Report < ::RollbarAPI
--- a/spec/support/deploy_api/update.rb
+++ b/spec/support/deploy_api/update.rb
@@ -1,4 +1,5 @@
 require 'rack/request'
+require "#{File.dirname(__FILE__)}/../rollbar_api.rb"
 
 module DeployAPI
   class Update < ::RollbarAPI


waltjones:
@scudco Thank you!

Issue #880, opened 2019-06-24
Rollbar triggers autoloading during initialization in Rails 6.0.0.rc1


pvande:
**Steps to Reproduce**
- Create a new app with Rails 6.0.0.rc1.
- Add `gem 'rollbar'` to the Gemfile.
- Run `bundle install`.
- Run `rails environment`.
- Run `cat log/development.log`.

**Expected Results**
The `development.log` file is empty.

**Actual Results**
The log file contains the following deprecation warning:

```
DEPRECATION WARNING: Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload ActionText::ContentHelper, for example,
the expected changes won't be reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <main> at /Users/
USER/Projects/rails-app-test/config/environment.rb:5)
```

**Analysis**
The autoloaded constants in question are coming from [this initializer](https://github.com/rails/rails/blob/v6.0.0.rc1/actiontext/lib/action_text/engine.rb#L32-L34) in Action Text.  This callback expects to wait until after ActionController::Base has been loaded before running.  This would naturally happen *after* the initialization process (say, after autoloading an application controller), but Rollbar's [`RailtieMixin`](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/plugins/rails/railtie_mixin.rb#L6-L8) immediately loads the `lib/rollbar/rake_tasks.rb`.  This file only defines a simple test task to verify that Rollbar is functioning, but in the process [creates a Rails controller](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/rake_tasks.rb#L143-L153) that descends from `ActionController::Base`.  When executed, this controller declaration automatically loads `ActionController::Base`, triggering the callbacks, and causing the helpers to become loaded – all before initialization has been completed.

One potential change that would remedy this fault and simultaneously reduce the startup time impact of the Rollbar gem would be to move the entire testing setup into its own file which is then required on-demand from the Rake task.

``` ruby
namespace :rollbar do
  desc 'Verify your gem installation by sending a test exception to Rollbar'
  task :test => [:environment] do
    require './rollbar_test'
    RollbarTest.run
[...]

PR #881, opened 2019-06-24
Move the verification code into a lazy-loaded file

pvande:
This will help improve load time for the library while reducing load-time side-effects.  In particular, this will avoid forcing ActionController::Base to load in Rails applications during initialization.

Proposed as a potential solution to #880.


lib/rollbar/rake_tasks.rb
```diff
@@ -1,154 +1,7 @@
...
-  require 'rack/mock'
-rescue LoadError
-  puts 'Cannot load rack/mock'
-end
-require 'logger'
-
 namespace :rollbar do
   desc 'Verify your gem installation by sending a test exception to Rollbar'
   task :test => [:environment] do
+    require './verify_setup'
```
waltjones: This didn't work for me when executing `rake rollbar:test` from a Rails app. It takes the path relative to the app rather than rollbar-gem, and the require fails.

waltjones:
@pvande Apologies I didn't see this PR sooner. 

https://github.com/rollbar/rollbar-gem/pull/882 includes updated rspec/travis for Rails 6.x, and since the reported issue has behavior specific to Rails 6.x, I wanted to get those updates in as well.

Thank you for the work on this, especially the very complete analysis.

pvande:
@waltjones No worries!  I just figured that since I'd already done the legwork I might as well put together a PR too.  Cheers!

Issue #906, opened 2019-10-21
Rollbar gem breaks Rails 6's new config defaults for ActiveJob


mrhead:
## Issue description

When you upgrade a Rails 5.2 app to Rails 6, then `rails app:update` creates `config/initializers/new_framework_defaults_6_0.rb` with new Rails 6 defaults which can be enabled one by one. One of them is:

```ruby
# Return false instead of self when enqueuing is aborted from a callback.
Rails.application.config.active_job.return_false_on_aborted_enqueue = true
```

This doesn't work with rollbar > 2.18.2 and the setting is ignored.

## Steps to reproduce

Please check [this](https://github.com/mrhead/rollbar-rails-defaults-bug) repository which simulates a Rails 6 application and contains a test which reproduces this issue.

## Workaround

Set `Rails.application.config.active_job.return_false_on_aborted_enqueue` in `config/application.rb` instead of `new_framework_defaults_6_0.rb`. Or, if your app is ready switch to new Rails 6 defaults then you can use `config.load_defaults 6.0` which works just fine.

## What is causing the issue?

I presume that the issue is caused by [this line](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/delay/active_job.rb#L5) which calls `::ActiveJob::Base` which causes `ActiveJob` to initialize sooner than it should. I think that the whole file should be wrapped in `ActiveSupport.on_load(:active_job)` instead.

Please check the following list of similar issues for more information:

- https://github.com/rails/rails/issues/23589
- https://github.com/plataformatec/devise/commit/c2c74b0a39238e7d997486814a1c8f75fdaf276f
- https://github.com/rubysherpas/paranoia/pull/371/files

waltjones:
@mrhead Thank you for the thorough analysis and the workaround.

PR #915, opened 2019-11-15
Allow ActiveJob to lazy load

waltjones:
Fixes: https://github.com/rollbar/rollbar-gem/issues/906

Rollbar plugins are eager loaded, which doesn't interfere with Rails ActiveJob settings, except where ActiveJob is included in `ActionMailer::DeliveryJob`. Eager loading here causes settings in initializers to not get picked up. Using `ActiveSupport.on_load` fixes this, but we take care to hook load of ActionMailer, not ActiveJob.

### DIFF
--- a/lib/rollbar/plugins/active_job.rb
+++ b/lib/rollbar/plugins/active_job.rb
@@ -14,5 +14,7 @@ module Rollbar
   end
 end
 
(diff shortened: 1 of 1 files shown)


mrhead:
Nice work @waltjones!

PR #925, opened 2019-12-19
Don't try and call ActiveSupport without checking if it's defined

steveh:
Non-Rails Ruby apps may not have ActiveSupport.

Perhaps a non-Rails build should be added to the Travis matrix?


waltjones:
Agreed. Thank you for the PR, and apologies for the inconvenience. I'll merge and release when it goes green.

PR #938, opened 2020-02-04
Rollbar pid files

Aristat:
Hi! We want to save the logs to different files for each puma workers, is it possible to add such a PR to the gem?

in addition:
+ small refactor old code
+ added functionality for rename `processed` files(for smooth file processing by our agent)

thanks!

### DIFF
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -19,7 +19,8 @@ module Rollbar
     attr_accessor :last_report
     attr_accessor :scope_object
 
-    @file_semaphore = Mutex.new
+    MUTEX = Mutex.new
+    EXTENSION_REGEXP = /.rollbar\z/.freeze
 
     def initialize(parent_notifier = nil, payload_options = nil, scope = nil)
       if parent_notifier
@@ -200,11 +201,11 @@ module Rollbar
     def process_item(item)
       if configuration.write_to_file
         if configuration.use_async
-          @file_semaphore.synchronize do
-            write_item(item)
+          MUTEX.synchronize do
+            do_write_item(item)
           end
         else
-          write_item(item)
+          do_write_item(item)
         end
       else
         send_item(item)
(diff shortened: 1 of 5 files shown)


waltjones:
@Aristat Thank you for the PR. Just a few questions.

lib/rollbar/notifier.rb
```diff
@@ -675,34 +676,41 @@ def ssl_verify_mode
...
-      if configuration.use_async
-        @file_semaphore.synchronize do
-          do_write_item(item)
-        end
-      else
-        do_write_item(item)
-      end
-    end
-
     def do_write_item(item)
```
waltjones: Is there a reason other than style for the method rename `write_item` -> `do_write_item`?

lib/rollbar/notifier.rb
```diff
@@ -675,34 +676,41 @@ def ssl_verify_mode
...
       log_info '[Rollbar] Writing item to file'
 
       body = item.dump
       return unless body
 
+      file_name = if configuration.files_with_pid_name_enabled
+                    configuration.filepath.gsub(EXTENSION_REGEXP, "_#{Process.pid}\\0")
```
waltjones: I'm not familiar with using the backslash zero in the filename here. Is this a convention?

lib/rollbar/notifier.rb
waltjones: I misread that. I see it now.

lib/rollbar/notifier.rb
Aristat: yeah, https://apidock.com/ruby/v2_6_3/String/gsub (see examples `Replacing with "\" and match`), it's just needed for replace `file_name.rollbar` to `file_name_12345.rollbar`
since in `filepath` includes extension and I have to do the replacement in this way(I specified that new flags only work if `.rollbar` is at the end, because it is hardcoded in a gem)

https://github.com/rollbar/rollbar-gem/blob/ce19357f2349d6d56426cc98341805ce8251a0dd/lib/rollbar/plugins/rails/railtie_mixin.rb#L17

waltjones: I see, it's a backreference for gsub. Thank you.

PR #941, opened 2020-02-06
Update file with sync operation

Aristat:
Need to add a cleanup file variable, after we renamed, but I don’t know how to do it better, just set it to `nil` or create a file?

Problem reproduced only without asyn logic(because we use one instance)
Without this fix, we will continue to write logs to the already renamed file

thanks!

### DIFF
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -709,6 +709,7 @@ module Rollbar
 
       new_file_name = file_name.gsub(EXTENSION_REGEXP, "_processed_#{time_now.to_i}\\0")
       File.rename(file, new_file_name)
+      @file = File.open(file_name, 'a')
     end
 
     def failsafe_reason(message, exception)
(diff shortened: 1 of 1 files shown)


waltjones:
You also need to close the previous file handle.

I should have asked before, but is there a reason you can't use logrotate (or equivalent)?

Aristat:
yes, need close, I will fix, thx.
if you mean https://github.com/logrotate/logrotate, we already use rollbar on many projects, but because of ddos - redis / sidekiq increases the load, and we want to write an agent that would read data from files so as not to remove rollbar from the project

waltjones:
Logrotate specifically manages rotation of log files. It can be used to do all of the opening, closing, and renaming of log files, including the log files from Rollbar.

Aristat:
Isn't collision possible with file flush there?(lost data or something else)?

waltjones:
With logrotate? There shouldn't be. That's exactly the use it was designed for, and has been used for years in large ops environments.

Aristat:
sounds good, we will try it too, thanks

PR #942, opened 2020-02-19
Handle error encoding string without failing the occurrence

waltjones:
Specific strings in an occurrence may contain encodings that cannot map to UTF-8. For example, the Vietnamese encoding Windows-1258 allows sequences that have no valid mapping in UTF-8.

When a mapping fails, it is better to provide a diagnostic error for the specific string than to fail the occurrence. Failsafe error occurrences don't show where the failed string mapping happened, and don't provide useful information about the original error. Preserving and sending the original error is an improvement on both points.

### DIFF
--- a/lib/rollbar/encoding/encoder.rb
+++ b/lib/rollbar/encoding/encoder.rb
@@ -30,6 +30,9 @@ module Rollbar
                         end
 
         object.is_a?(Symbol) ? encoded_value.to_sym : encoded_value
+      rescue StandardError => e
+        # If encoding fails for any reason, replace the string with a diagnostic error.
+        "error encoding string: #{e.class}: #{e.message}"
       end
 
       private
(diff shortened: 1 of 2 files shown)


lib/rollbar/encoding/encoder.rb
```diff
@@ -30,6 +30,9 @@ def encode
                         end
 
         object.is_a?(Symbol) ? encoded_value.to_sym : encoded_value
+      rescue StandardError => e
+        # If encoding fails for any reason, replace the string with a diagnostic error.
+        "error encoding string: #{e.class}: #{e.message}"
```
bf4, line 35: I have a really robust solution in the [EncodedString gem](https://github.com/bf4/encoded_string/blob/04e4f246d53b2ccf77f933addbf56648aa470783/lib/encoded_string.rb) which I extracted from some work in rspec `¯\_(ツ)_/¯`

use it as a reference or rely on it. Either way.

Issue #943, opened 2020-02-24
[Rollbar] Error processing the item: NameError, uninitialized constant JSON.


patricksimpson:
Running within a docker container, using Ruby version 2.7.0. 

`ruby rollbar-test.rb`

`Error processing the item: NameError, uninitialized constant JSON.`

rollbar-test.rb: 
```
require 'rollbar'

Rollbar.configure do |config|
  config.access_token = '...' #filtered out token
end

Rollbar.info("Script ran successfully")
```

Added `require 'json'` to the top of that file and it works. 
Shouldn't this be a require dependency already? 

I am not running this with `rake`.

waltjones:
@patricksimpson Thank you for the report, and the test.

I'm not able to reproduce this on 2.5.3, 2.6.3 or 2.7.0. I'm using exactly your test file, with a valid access token. No bundler, local rollbar-gem (2.24.0).

I don't imagine the docker container would make a difference. I'm thinking there must be some other difference between my test and yours.

patricksimpson:
@waltjones thanks for the reply. Yeah, this is strange for sure. Locally, (MacOS) using the same ruby version, it works fine. Only in this docker container. I am going to try a few different images and see if that makes a difference. Sorry I will close this out if I can't find anything other than my lonely custom image.

patricksimpson:
@waltjones Ok using the 2.7.0 Docker:
`FROM ruby:slim-buster `
https://hub.docker.com/layers/ruby/library/ruby/slim-buster/images/sha256-33d1e540c826b75d69f841da3321fe1958735f7233d7818cf1099dd93ea191f9?context=explore

Doing a `bundle install` and running the above file without the `require 'json'` 

`E, [2020-02-25T04:07:13.209017 #1] ERROR -- : [Rollbar] Sending failsafe response due to NameError: "uninitialized constant JSON" in /usr/local/bundle/gems/rollbar-2.24.0/lib/rollbar/json.rb:11:in `block in dump': error in process_item`

waltjones:
@patricksimpson Thank you for the followup. I'm able to reproduce this with your linked docker image, and with non-slim images as well.

I'll get a fix into the next release.

PR #944, opened 2020-02-28
Require json necessary in some environments

waltjones:
Fixes: https://github.com/rollbar/rollbar-gem/issues/943

The missing require doesn't fail on rvm installed rubies, even when testing with a minimal Ruby script and no other dependencies. It does fail on these docker images however: https://hub.docker.com/_/ruby

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -121,6 +121,8 @@ module Rollbar
       end
 
       def config_js_tag(env)
+        require 'json'
+
         js_config = Rollbar::Util.deep_copy(config[:options])
 
         add_person_data(js_config, env)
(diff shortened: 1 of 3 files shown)


lib/rollbar/middleware/js.rb
```diff
@@ -121,6 +121,8 @@ def close_old_response(response)
       end
 
       def config_js_tag(env)
+        require 'json'
```
vselvarajijay, line 124: just curious, do we need to keep a require 'json' inside the method? why not with all the other requires?

waltjones: In Ruby, this puts the require only in the scope of that method. (The only method where we call `JSON.generate()`.) This allows the rest of the code to work without the dependency.

PR #954, opened 2020-04-24
Update rollbar.js 2.15.2

waltjones:
Update rollbar.js snippet from 2.12.3 to 2.15.2.

https://github.com/rollbar/rollbar.js/releases


vselvarajijay:
If the unit tests pass - this is good to me

PR #958, opened 2020-05-07
support report_level configuration just like rollbar.js

v-kumar:
Disable reporting based on `report_level` configuration, similar to https://docs.rollbar.com/docs/rollbarjs-configuration-reference


waltjones:
@v-kumar Thank you for the PR. We're trying not to add anything new to spec/rollbar_bc_spec.rb, and  instead the specs should go in spec/rollbar_spec.rb.

I'd suggest adding either before or after the `exception_level_filters` specs.
https://github.com/rollbar/rollbar-gem/blob/master/spec/rollbar_spec.rb#L794_L922

PR #961, opened 2020-05-12
Fix Rollbar JS middleware truncating response body

### LINKED ISSUE #960: Rollbar JS middleware truncates response body
## Problem

I have a Rails application using `rollbar` and [`split`](https://github.com/splitrb/split) gems. Split has a dashboard written in Sinatra, which you can mount in your Rails routes.

When I try to access Split dashboard, I see a blank page. After some debugging I tracked down the problem might be in `Rollbar::Middleware::Js`, which is replacing the response body while not updating the `Content-Length` header.

I am not entirely sure why this is happening only to mounted engines/rack apps (didn't have time to look deeper into it), once the main application responses are not getting truncated, probably because at that point there's no `Content-Length` header yet and this header is added somewhere else.

## How to reproduce

I created a minimal Rails application to reproduce the issue. See `config/initializers/rollbar.rb` for more information.

[rollbar_issue.tar.gz](https://github.com/rollbar/rollbar-gem/files/4615846/rollbar_issue.tar.gz)

## Solution

Applying the following patch seems to solve the issue. I can open a PR with a test case and this patch if you guys are happy with the solution.

```ruby
module Rollbar
  module Middleware
    class Js
      def build_response(env, app_result, response_string)
        return app_result unless response_string

        env[JS_IS_INJECTED_KEY] = true

        status, headers, = app_result
        headers['Content-Length'] = response_string.bytesize if headers.key?('Content-Length')

        response = ::Rack::Response.new(response_string, status, headers)
        finished = response.finish

        # Rack < 2.x Response#finish returns self in array[2]. Rack >= 2.x returns self.body.
        # Always return with the response object here regardless of rack version.
        finished[2] = response
        finished
      end
    end
  end
end
```

wpolicarpo:
The JS middleware injects Rollbar JS snippet into the response body if not already added.

If for some reason the `Content-Length` header was previously set by another middleware, we also need to update that header to reflect the new response body size.

Fixes #960.

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -80,8 +80,11 @@ module Rollbar
         return app_result unless response_string
 
         env[JS_IS_INJECTED_KEY] = true
-        response = ::Rack::Response.new(response_string, app_result[0],
-                                        app_result[1])
+
+        status, headers, = app_result
+        headers['Content-Length'] = response_string.bytesize.to_s if headers.key?('Content-Length')
+
+        response = ::Rack::Response.new(response_string, status, headers)
 
         finished = response.finish
 
--- a/spec/rollbar/middleware/js_spec.rb
+++ b/spec/rollbar/middleware/js_spec.rb
@@ -179,7 +179,7 @@ describe Rollbar::Middleware::Js do
         end
       end
 
-      context 'having a html 200 resposne with meta charset tag' do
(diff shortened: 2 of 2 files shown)


waltjones:
@wpolicarpo Thank you!

waltjones:
https://github.com/rollbar/rollbar-gem/releases/tag/v2.26.0

PR #966, opened 2020-05-27
Remove access_token from item payload

### LINKED ISSUE #965: access_token is present in logs after error
When there is an error talking to rollbar, the following is logged locally including the access_token in plain text.

```
2.6.3 :010 > Rollbar.debug("Running Script")
I, [2020-05-27T14:29:00.099017 #49157]  INFO -- : [Rollbar] Scheduling item
I, [2020-05-27T14:29:00.099067 #49157]  INFO -- : [Rollbar] Sending item
I, [2020-05-27T14:29:00.099160 #49157]  INFO -- : [Rollbar] Sending json
E, [2020-05-27T14:29:00.101092 #49157] ERROR -- : [Rollbar] Error processing the item: Errno::ECONNREFUSED, Failed to open TCP connection to api.rollbar.com:443 (Connection refused - connect(2) for "api.rollbar.com" port 443). Item: {"access_token"=>"AAAAAAAAAAAAAAAAAAAAA", "data"=>{:timestamp=>1590614940, :environment=>"unspecified", :level=>"debug", :language=>"ruby", :framework=>"Plain", :server=>{:host=>"localhost", :pid=>49157}, :notifier=>{:name=>"rollbar-gem", :version=>"2.25.0", :configured_options=>{:access_token=>"******"}}, :body=>{:message=>{:body=>"Running Script"}}, :uuid=>"e17a869e-06c6-456b-a12a-963deb6b13d7"}}
E, [2020-05-27T14:29:00.101141 #49157] ERROR -- : [Rollbar] Reporting internal error encountered while sending data to Rollbar.
```

Note that "AAAAAAAAAAAAAAAAAAAAA" is the access token which is not redacted in the log. When it's specified in the options it is redacted to "*******"

swangs:
Resolves #965 

After some investigation, it looks like the access_token in the item payload is no longer used in this gem. Notifier requests are now sent with a "X-Rollbar-Access-Token" header. Removing the token from the payload will also remove it from logs. 

The RollbarAPI stub is still expecting the payload to include an access token so I've changed that expectation to look at the header. The Report stub inherits from RollbarAPI but has a different structure now. It's been updated to validate against the access_token in its request body rather than to super RollbarAPI's behavior. 

One more change is that the payload saved to file will also no longer include the access_token. I'm not sure what consequences this might have. If you know that this is a necessary field, we can to make changes to include it here.

### DIFF
--- a/lib/rollbar/item.rb
+++ b/lib/rollbar/item.rb
@@ -65,7 +65,6 @@ module Rollbar
     def build
       data = build_data
       self.payload = {
-        'access_token' => configuration.access_token,
         'data' => data
       }
 
--- a/lib/rollbar/notifier.rb
+++ b/lib/rollbar/notifier.rb
@@ -290,7 +290,6 @@ module Rollbar
       }
 
       failsafe_payload = {
-        'access_token' => configuration.access_token,
         'data' => failsafe_data
       }
 
--- a/spec/rollbar/item_spec.rb
+++ b/spec/rollbar/item_spec.rb
@@ -53,7 +53,7 @@ describe Rollbar::Item do
       end
 
       it 'should have the correct root-level keys' do
-        payload.keys.should match_array(['access_token', 'data'])
+        payload.keys.should match_array(['data'])
       end
 
       it 'should have the correct data keys' do
@@ -566,7 +566,6 @@ describe Rollbar::Item do
       context 'with mutation in payload' do
         let(:new_payload) do
           {
-            'access_token' => configuration.access_token,
             'data' => {
             }
           }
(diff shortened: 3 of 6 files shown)


waltjones:
@swangs Thank you. Will get a release out as soon as possible.

grosser:
this breaks sending for us
```
{"severity":"INFO","message":"[Rollbar] Sending json"}
irb(main):002:0> {"severity":"WARN","message":"[Rollbar] Got unexpected status code from Rollbar api: 422"}
{"severity":"INFO","message":"[Rollbar] Response: {\n  \"err\": 1,\n  \"message\": \"Invalid format. access_token is missing and it is required.\"\n}"}
```

... can you add some instructions to UPGRADING.md ?

russell:
The fix for this is to add the `use_payload_access_token` to disable the new behaviour, so you see the 422 like above try this

```
Rollbar.configure do |config|
    config.use_payload_access_token = true
end
```

PR #968, opened 2020-06-09
Support any object as extra data when notifying errors

bagedevimo:
Currently the extra_data support in rollbar-ruby relies on the object
being of a specific set of types. This is a bit uncessary, as we do
support hashes. So instead of limiting the types by their class name,
lets instead check the response to `to_h` (which always returns hash)
and then treat the result as a hash.

This enables support for a larger number of extra_data objects being
passed in, including ActiveModel::Errors, the original object that was
being passed in at my work.

### DIFF
--- a/spec/rollbar_spec.rb
+++ b/spec/rollbar_spec.rb
@@ -190,6 +190,29 @@ describe Rollbar do
         notifier.log('error', exception, extra_data, 'exception description')
       end
 
+      context 'when the extra data is something that supports to_h' do
+        # This is faking ActiveModel::Errors, which is the use case this feature was
+        # originally added in order to support.
+        # Generated from a prod-app by doing the following, where `opn` is some Rails/ActiveRecord model
+        # irb(main):007:0> opn.errors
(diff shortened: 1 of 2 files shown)


spec/rollbar_spec.rb
```diff
@@ -190,6 +190,29 @@
...
+        # irb(main):008:0> opn.errors.to_h
+        # => {:payload=>"can't be blank"}
+        FakeActiveModelErrors = Struct.new(:messages) do
+          def to_h
+            messages.to_h[:messages]
+          end
+        end
+
+        let(:extra_data) { FakeActiveModelErrors.new(messages: { "payload": "can't be blank" }) }
+        let(:extra_data_hash) { { "payload": "can't be blank" } }
```
waltjones, line 208: Hashrocket syntax is required in this project for back-compatibility, and single quotes for plain (non-interpolated) strings.
e.g.
`{ :payload =>  'can't be blank' }`

waltjones:
@bagedevimo Thank you for the PR. A release is coming up in the next couple days. I'd be happy to merge this change if you can update the hash syntax in the spec.

PR #969, opened 2020-06-16
Send original error info with failsafe

waltjones:
When sending a failsafe error, this adds information about the original error that otherwise wouldn't be present.

This PR also uses `process_item` instead of `schedule_item`, to keep from putting the failsafe error back into an async job queue.

### DIFF
--- a/lib/rollbar/item.rb
+++ b/lib/rollbar/item.rb
@@ -136,11 +136,18 @@ module Rollbar
       uuid = stringified_payload['data']['uuid']
       host = stringified_payload['data'].fetch('server', {})['host']
 
+      original_error = {
+        :message => message,
+        :exception => exception,
+        :configuration => configuration,
+        :uuid => uuid,
+        :host => host
+      }
+
       notifier.send_failsafe(
         too_large_payload_string(attempts),
         nil,
-        uuid,
-        host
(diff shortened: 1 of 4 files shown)


lib/rollbar/item.rb
```diff
@@ -136,11 +136,18 @@ def handle_too_large_payload(stringified_payload, final_payload, attempts)
       uuid = stringified_payload['data']['uuid']
       host = stringified_payload['data'].fetch('server', {})['host']
 
+      original_error = {
+        :message => message,
+        :exception => exception,
+        :configuration => configuration,
+        :uuid => uuid,
+        :host => host
+      }
```
brianr, line 145: Is there any possibility that including this extra data could cause the failsafe to fail? e.g. could there be an error in serialization, could the payload become too large...

(IIRC that's why this data was omitted from the failsafe before.)

waltjones: The intent is that everything that might fail is wrapped in a rescue, and if it fails we send the error message string instead. Payload size seems like an unlikely issue for the data that is sent, but I've added a commit that does truncate the unknown-length strings, and pre-serializes the options object to ensure only simple strings are sent for final serialization.

Issue #972, opened 2020-06-24
Capistrano 2: rollbar reports incorrect revision


mrchucho:
I know Capistrano 2 has fallen out of fashion, so I'm sure this is a long shot. But rollbar's Capistrano 2 deploy integration reports the wrong revision when deploying. It looks like rollbar uses Capistrano's [current_revision](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/capistrano.rb#L38) to report the revision being deployed. Unfortunately `current_revision` is actually the _currently deployed revision_ not the revision being deployed. In other words, rollbar reports a deployment of what is already deployed not what is being deployed.

waltjones:
Released in v2.26.1 https://github.com/rollbar/rollbar-gem/releases/tag/v2.26.1

PR #973, opened 2020-06-24
Use real revision

mrchucho:
Fixes: https://github.com/rollbar/rollbar-gem/issues/972

I.e. use what we're deploying not what is deployed


mrchucho:
This should resolve #972

waltjones:
@mrchucho :+1: Thank you. Will go in the next release.

PR #978, opened 2020-07-11
Recreate REQUEST_URI using Rack SPEC-compliant variables

joelvh:
Turns out `REQUEST_URI` is not in the Rack SPEC, and [the recommendation](https://github.com/rack/rack/issues/1342#issuecomment-573926795) is to reproduce the same value value using SPEC-compliant variables. [`Rack::Request#fullpath`](https://github.com/rack/rack/blob/e15eafc28170ea759407a031b68d690fbd83d3be/lib/rack/request.rb#L515) builds the equivalent of `REQUEST_URI`, but we don't want to add a dependency on `Rack` directly.

The update uses Rack-compliant variables.

### DIFF
--- a/lib/rollbar/request_data_extractor.rb
+++ b/lib/rollbar/request_data_extractor.rb
@@ -131,6 +131,7 @@ module Rollbar
       host = host.split(',').first.strip unless host.empty?
 
       path = env['ORIGINAL_FULLPATH'] || env['REQUEST_URI']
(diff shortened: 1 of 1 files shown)


waltjones:
@joelvh Thank you for the PR.

joelvh:
No problem @waltjones!

waltjones:
Released in v2.26.1 https://github.com/rollbar/rollbar-gem/releases/tag/v2.26.1

Issue #989, opened 2020-08-07
Rollbar is using URI.escape that is already obsolete in ruby v2.7


pangtongs:
Rollbar is using URI.escape that is already obsolete in ruby v2.7
and giving me some noise in error debugging
```
/usr/local/bundle/gems/rollbar-2.26.0/lib/rollbar/scrubbers/url.rb:73: warning: URI.escape is obsolete
```

waltjones:
@pangtongs Thank you for the report. Will get this updated soon.

waltjones:
Released in v2.27.1

PR #993, opened 2020-08-27
Remove deprecated URI.escape

waltjones:
## Description of the change

Removes the deprecated `URI.escape`, which has been deprecated for a long time, but the warnings are now more visible in Ruby 2.7.

The previous code had ended up performing several transforms in order to unencode square brackets while keeping everything else correctly encoded. Condensed, it was effectively like this:
`URI.escape(CGI.unescape(URI.encode_www_form(params)))`

This is all necessary because `URI.encode_www_form` percent encodes square brackets, but we want them unencoded.

This PR refactors and takes a simpler approach. As a side note, the debate about whether unencoded brackets should be allowed in URLs is not as relevant here since these are only used as debugging data in Rollbar payloads. They aren't being used to perform actual http requests.

The original URL may or may not have been percent encoded. The previous code didn't preserve the original state, and neither does this PR. The unencoded brackets are more readable, and is what everyone has been expecting and getting in their Rollbar dashboard, so we keep it that way.

## Type of change

## Related issues

Fixes: https://github.com/rollbar/rollbar-gem/issues/989
ch71576

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/lib/rollbar/scrubbers/url.rb
+++ b/lib/rollbar/scrubbers/url.rb
@@ -67,10 +67,7 @@ module Rollbar
 
         params = decode_www_form(query)
 
-        encoded_query = encode_www_form(filter_query_params(params, regex, randomize_scrub_length, scrub_all, whitelist))
-
-        # We want this to rebuild array params like foo[]=1&foo[]=2
-        URI.escape(CGI.unescape(encoded_query))
+        encode_www_form(filter_query_params(params, regex, randomize_scrub_length, scrub_all, whitelist))
       end
 
       def decode_www_form(query)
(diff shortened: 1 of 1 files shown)


lib/rollbar/scrubbers/url.rb
```diff
@@ -67,18 +67,23 @@ def filter_query(query, regex, randomize_scrub_length, scrub_all, whitelist)
...
-        URI.encode_www_form(params)
+        restore_square_brackets(URI.encode_www_form(params))
+      end
+
+      def restore_square_brackets(query)
+        # We want this to rebuild array params like foo[]=1&foo[]=2
+        #
+        # URI.encode_www_form follows the spec at https://url.spec.whatwg.org/#concept-urlencoded-serializer
+        # and percent encodes square brackets. Here we change them back.
+        query.gsub('%5B', '[').gsub('%5D', ']')
```
letaniaferreira, line 86: Simple but interesting stuff. I didn't know about gsub before.

PR #1004, opened 2020-11-04
Restore the option to put the token in the payload if needed

waltjones:
### Background

In the specific issue report, the access token is changed dynamically in the config in order to send reports to multiple Rollbar projects. An async/delayed sender is used, which doesn't have the dynamically updated config. Therefore all reports are sent using the access token in the delayed sender config.

The above scenario used to work because the token was added to the payload before queuing. Recent versions of rollbar-gem don't add the token to the payload and only add it as an http request header. (See https://github.com/rollbar/rollbar-gem/pull/966) The header is added by the delayed sender process, which uses the access token in its config.

The current interface for async senders only passes the payload and no other arguments, and the sender process doesn't have access to any other context. Even if this could be augmented in a back compatible way for the built-in senders, custom senders would still be affected by this.

### Solution

An option (`config.use_payload_access_token = true`) is added to use the payload instead of the http header for the access token. The http header is specifically omitted in this case, to ensure the Rollbar API uses the intended token when they differ. When the config option is set, the token is sent in the payload regardless of whether the report is async or not.

This can be deprecated and removed on a future major release where an improved interface for async jobs is available. 

## Type of change

## Related issues

Fixes ch74796

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -75,6 +75,7 @@ module Rollbar
     attr_accessor :files_processed_enabled
     attr_accessor :files_processed_duration # seconds
     attr_accessor :files_processed_size # bytes
+    attr_accessor :use_payload_access_token
 
     attr_reader :project_gem_paths
     attr_accessor :configured_options
@@ -167,6 +168,7 @@ module Rollbar
       @files_processed_enabled = false
       @files_processed_duration = 60
       @files_processed_size = 5 * 1000 * 1000
+      @use_payload_access_token = false
 
       @configured_options = ConfiguredOptions.new(self)
     end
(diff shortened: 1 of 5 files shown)


letaniaferreira (approved):
LGTM!

grosser:
found this workaround in multiple of our projects, some UPGRADING.md instructions especially on how to actually fix and not keep sending the token would be great

waltjones:
@grosser Can you elaborate on how you're sending that causes you to hit this issue? It will help ensure that case is covered in the doc.

grosser:
```
Rollbar.configure do |config|
  config.environment = ENV.fetch("RAILS_ENV")
  config.access_token = ENV.fetch("ROLLBAR_ACCESS_TOKEN")
  config.endpoint = ENV.fetch('ROLLBAR_URL') + '/api/1/item/'
  config.web_base = ENV.fetch('ROLLBAR_URL')
  config.use_thread
  config.code_version = ENV.fetch('TAG')
end
```

PR #1010, opened 2020-11-10
Remove skipping CSP nonce on `unsafe-inline`

### LINKED ISSUE #1009: CSP incorrectly skips adding nonce when `unsafe-inline` is set
Rollbar's support for Rails's CSP nonce functionality intentionally skips adding a nonce to the script tag if the `script-src` directive includes `unsafe-inline`: https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/js.rb#L186

The rationale for this is described in the relevant spec:
```ruby
# Browser behavior is undefined when unsafe_inline and the nonce are both present.
# The app should never set both, but if they do, our best behavior is to not use the nonce.
```

However, using both a nonce **and** `unsafe-inline` at the same time is perfectly valid (and indeed sensible) behaviour. It allows the app to maintain some level of backwards compatibility with browsers that support CSP1 but not CSP2.

See [CSP `script-src` on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src):
> Specifying nonce makes a modern browser ignore 'unsafe-inline' which could still be set for older browsers without nonce support.

[Google's CSP checker](https://csp-evaluator.withgoogle.com/) will also issue a warning if a nonce is specified without `unsafe-inline`.

I was going to PR this but figured it might throw up a fair few headaches around changing existing behaviour so wanted to start a conversation first! 😃

csutter:
# Description of the change
The current implementation for Rails's CSP nonce functionality
intentionally skips adding a nonce to the script tag if the script-src
directive includes `unsafe-inline`. However, using both a nonce and
unsafe-inline at the same time is perfectly valid (and indeed sensible)
behaviour. It allows the app to maintain some level of backwards
compatibility with browsers that support CSP1 but not CSP2, c.f.:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

## Type of change

## Related issues

Fixes #1009.

### Development

- [x] Lint rules pass locally

### Code review 

- [ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Issue from task tracker has a link to this pull request

### DIFF
--- a/lib/rollbar/middleware/js.rb
+++ b/lib/rollbar/middleware/js.rb
@@ -183,7 +183,6 @@ module Rollbar
         req.respond_to?(:content_security_policy) &&
           req.content_security_policy &&
           req.content_security_policy.directives['script-src'] &&
-          !req.content_security_policy.directives['script-src'].include?("'unsafe-inline'") &&
           req.content_security_policy_nonce
       end
 
@@ -224,15 +223,11 @@ module Rollbar
         end
 
         def csp_needs_nonce?(csp)
-          !opt_out?(csp) && !unsafe_inline?(csp)
+          !opt_out?(csp)
         end
 
         def opt_out?(_csp)
           raise NotImplementedError
-        end
-
-        def unsafe_inline?(csp)
-          csp[:script_src].to_a.include?("'unsafe-inline'")
         end
       end
 
--- a/spec/rollbar/middleware/js_spec.rb
+++ b/spec/rollbar/middleware/js_spec.rb
@@ -17,20 +17,6 @@ shared_examples 'secure_headers' do
     expect(new_body).to include(snippet)
   end
 
-  it 'renders the snippet in the response without nonce if SecureHeaders script_src includes \'unsafe-inline\'' do
-    SecureHeadersMocks::CSP.config = {
-      :opt_out? => false,
-      :script_src => %w['unsafe-inline'] # rubocop:disable Lint/PercentStringArray
-    }
-
-    _, _, response = subject.call(env)
-    new_body = response.body.join
-
-    expect(new_body).to include('<script type="text/javascript">')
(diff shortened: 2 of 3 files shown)


csutter:
FWIW, I can't think of anything this PR might break - my digging through the spec suggested combining a nonce and `unsafe-inline` is definitely fine (see #1009).

Might still be worth someone else thinking about it too though 😅

csutter:
Hiya @waltjones – just wanted to cheekily chase and ask when you expect this to be merged 😄

waltjones (approved):
Thank you for the PR, and for the explanation.

waltjones:
Released in v3.1.2. Apologies for the delay.

pcasaretto:
I belive this update did in fact break our app.
The end result was that SecureHeaders added the nonce directive to the enforcing CSP header, at which point the browser ignored the `unsafe-inline` directive (as it should). This caused all inline script tags (everything but the rollbar script tag itself) without nonces to be ignored.
I do belive this behavior is a bug with secure headers itself, and I'm currenly [discussing it with the author](https://github.com/github/secure_headers/issues/470). But because of that, I believe this was a breaking change.

waltjones:
@pcasaretto Thank you for reporting this. I took a close look today, read your linked issue, and reviewed the CSP1 and CSP2 specs. I don't think reverting this works out well, as there are then a new set of issues. I'll work on a PR and link it here. Please feel free to give feedback there.

By the way, it looks like the Rails implementation for CSP does much the same as secure_headers. Just generating a nonce will set the header value. https://api.rubyonrails.org/classes/ActionDispatch/ContentSecurityPolicy/Request.html#method-i-content_security_policy_nonce

PR #1044, opened 2021-05-26
Rubocop: Bump version and apply safe autocorrect updates

waltjones:
## Description of the change

Rubocop has been pinned at version 0.93 for a long time, and also fell out of CI when Codacy was disabled for all SDKs. Most of the changes in this current round of PRs can be attributed to three factors:
* Rule changes introduced since 0.93
* Enforcing line length, which had never been enforced before
* Rollbar-gem was never fully green in the past, though nearly green except for method complexity.

Bringing rubocop up to date will be done in a series of PRs to help ease the code review burden, but also to get the changes that touch the most files (this PR) merged as soon as possible. Since this PR is limited to autocorrect fixes, review should be mostly mechanical and there hopefully shouldn't be a high cognitive load.

As usual, separate commits are used to help the reviewer.

* Updates rubocop from 0.93.0 to 1.15.0.
* TargetRubyVersion: 2.5 (This is the minimum allowed for current rubocop.)
* enable Layout/LineLength (max length 90)
* disable Style/RedundantBegin (Ruby < 2.5 needs begin/end inside blocks.)
* disable Layout/HeredocIndentation (For Ruby < 2.3. See below.)
* disable Lint/SendWithMixinArgument (Ruby 2.0 still has private `Object.include`)
* rubocop -a,  apply safe autocorrect changes

### TargetRubyVersion
Current rubocop only allows a minimum of 2.5 here, but we need to support 2.0 and up. We set this to the lowest allowed, and will need to modify specific rules as needed for differences in lower versions.

### Disable Style/RedundantBegin
The options for Layout/HeredocIndentation include modern squiggly syntax, and polyfill kinds of solutions, but not reliably the old syntax. Therefore we disable this rule for as long as we support < 2.3.

### Line length
The previous settings didn't only allow long lines. Frequently it _forced_ long lines because rubocop often prefers single line syntax for things if the line length limit allows it. Therefore many of the refactored long lines existed to prevent rubocop violations.

Upcoming PRs will address the following:
* Fix remaining rubocop issues.
* Reorganize cops in rubocop.yml. (for example, use lexicographic order of rules.)
* Remove obsolete Ruby 1.9 requirements
* Enable CI workflow
* Review performance cops, and update as needed

## Type of change
- [x] Style refactor

## Related issues

ch85855

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/lib/rails/rollbar_runner.rb
+++ b/lib/rails/rollbar_runner.rb
@@ -49,7 +49,7 @@ module Rails
     def legacy_runner
       string_to_eval = File.read(runner_path)
 
-      ::Rails.module_eval(<<-EOL, __FILE__, __LINE__ + 2)
+      ::Rails.module_eval(<<-EOL, __FILE__, __LINE__ + 1)
           #{string_to_eval}
       EOL
     end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,7 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'gemfiles/vendor/**/*'
+  TargetRubyVersion: 2.5 # This is the minimum allowed for current rubocop
 
 Style/HashSyntax:
   EnforcedStyle: hash_rockets
@@ -26,6 +27,10 @@ Style/Lambda:
   Enabled: false
 
 Style/EachWithObject:
+  Enabled: false
+
+Style/RedundantBegin:
+  # Ruby < 2.5 needs begin/end inside blocks when using rescue
   Enabled: false
 
 Metrics/BlockLength:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,8 +13,8 @@ Style/HashSyntax:
 Style/SymbolArray:
   Enabled: false # %i[] syntax isn't 1.9.x compatible
 
-Metrics/LineLength:
-  Max: 1000
+Layout/LineLength:
+  Max: 90
 
 Metrics/MethodLength:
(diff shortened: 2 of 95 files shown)


lib/rails/rollbar_runner.rb
```diff
@@ -49,7 +49,7 @@ def eval_runner
     def legacy_runner
       string_to_eval = File.read(runner_path)
 
-      ::Rails.module_eval(<<-EOL, __FILE__, __LINE__ + 2)
+      ::Rails.module_eval(<<-EOL, __FILE__, __LINE__ + 1)
```
diegov, line 52: Not a blocker but there's enough indirection here that I don't fully know why we're adding 1 to `__LINE__`. I assume it is set to 1 to match the location of `#{string_to_eval}`? If that's the case, was the 2 a bug?

This might benefit from a small comment.

diegov (approved):
:hot_face: :+1:

lib/rails/rollbar_runner.rb
waltjones: Great question. It sets location information in the event of an error with backtrace. The `2` seems like it was a bug, though a minor one. 

Yes this looks pretty esoteric, and I think what looks unexpected isn't the file/line args for `module_eval`. After all, those can be looked up quickly if you don't know how `module_eval` works. IMO it's the opening a heredoc in the first arg and finishing it on the following lines! Maybe a comment is deserved here.

PR #1048, opened 2021-06-04
Fix line length offenses

waltjones:
## Description of the change

This PR addresses all remaining line length offenses.

For literal strings, line continuation (`\`) is used by default. Ruby has several techniques for concatenation, but line continuation allows the parser to treat multiple lines as one literal string. This is more performant than concatenation and readability is improved as well.

In some cases, if a comment or test label should be more concise, the line length is solved by improving the text. I'm sure there are even more cases where this can be done, but I took the opportunity where it was obvious to me.

There are multiple places where "https://github.com" has been removed from a code reference in a comment, since it has been determined in earlier PRs that comments should not specifically reference Github URLs.

Long regular expressions are refactored to use the /x modifier. Other techniques are possible in Ruby, but this should be the most familiar to most people.

 Note: This PR also adds rubocop exclusions for spec/tmp and spec/dummyapp. 

## Type of change
- [x] Style

## Related issues

ch86350

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/lib/rollbar/configuration.rb
+++ b/lib/rollbar/configuration.rb
@@ -4,10 +4,79 @@ module Rollbar
   class Configuration
     SEND_EXTRA_FRAME_DATA_OPTIONS = [:none, :app, :all].freeze
 
-    attr_accessor :access_token, :async_handler, :branch, :capture_uncaught,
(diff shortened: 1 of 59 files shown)


spec/rollbar/item_spec.rb
```diff
@@ -291,7 +297,7 @@
       context 'with no exception' do
         let(:exception) { nil }
 
-        it 'should build a message body when no exception is passed in' do
+        it 'should build a message body' do
```
waltjones, line 300: This, and other examples like this one, already have these when/with modifiers in the enclosing context label. (At line 297 in this case.)

lib/rollbar/configuration.rb
```diff
@@ -143,7 +212,8 @@ def merge!(options)
     def use_active_job(options = {})
       require 'rollbar/delay/active_job'
 
-      Rollbar::Delay::ActiveJob.queue_as(options[:queue] || Rollbar::Delay::ActiveJob.default_queue_name)
+      Rollbar::Delay::ActiveJob.queue_as(options[:queue] ||
+      Rollbar::Delay::ActiveJob.default_queue_name)
```
diegov, line 216: I think some indentation might be missing here

lib/rollbar/notifier.rb
```diff
@@ -203,7 +203,9 @@ def critical(*args)
 
     def enabled?
       # Require access_token so we don't try to send events when unconfigured.
-      configuration.enabled && configuration.access_token && !configuration.access_token.empty?
+      configuration.enabled &&
+      configuration.access_token &&
```
diegov, line 207: I don't know what's common in Ruby, but I've always seen this type of multi-line statements with some indentation after the first line.

diegov (approved):

lib/rollbar/configuration.rb
waltjones: Rubocop's default setting is 'aligned', which is what we see above. Rubocop also allows 'indented', which would need to be an override in the rubocop config, otherwise indenting will produce a violation.

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineOperationIndentation

lib/rollbar/notifier.rb
waltjones: While I think this formatting is more common in Ruby, there are definitely projects that indent multi-line expressions. As noted above, we could choose to follow the other convention and update the rubocop config to match.

PR #1056, opened 2021-06-25
Resolve rubocop Style/CaseLikeIf violations

waltjones:
## Description of the change

Rubocop flags with Style/CaseLikeIf when if/elsif syntax would be better as case/when. The case/when syntax is generally more compact and readable, and this PR updates the violations to use case/when.

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/CaseLikeIf

## Type of change
- [x] Style

## Related issues

ch87751

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/lib/rollbar/exception_reporter.rb
+++ b/lib/rollbar/exception_reporter.rb
@@ -7,16 +7,17 @@ module Rollbar
 
       exception_data = exception_data(exception)
 
-      if exception_data.is_a?(Hash)
+      case exception_data
+      when Hash
         env['rollbar.exception_uuid'] = exception_data[:uuid]
         Rollbar.log_debug(
           "[Rollbar] Exception uuid saved in env: #{exception_data[:uuid]}"
         )
-      elsif exception_data == 'disabled'
+      when 'disabled'
         Rollbar.log_debug(
           '[Rollbar] Exception not reported because Rollbar is disabled'
         )
-      elsif exception_data == 'ignored'
+      when 'ignored'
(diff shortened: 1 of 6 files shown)


lib/rollbar/exception_reporter.rb
```diff
@@ -7,16 +7,17 @@ def report_exception_to_rollbar(env, exception)
...
 
-      if exception_data.is_a?(Hash)
+      case exception_data
+      when Hash
         env['rollbar.exception_uuid'] = exception_data[:uuid]
         Rollbar.log_debug(
           "[Rollbar] Exception uuid saved in env: #{exception_data[:uuid]}"
         )
-      elsif exception_data == 'disabled'
+      when 'disabled'
```
bxsx, line 16: This change is not 100% equivalent as it expresses `exception_data === 'disabled'`. It's probably still correct, but just want to point it out. In case it requires updating, there are few more similar changes in this PR.

bxsx (approved):
LGTM. I am leaving one ​comment just for double checking.

lib/rollbar/exception_reporter.rb
waltjones: Thanks for the comment. Strings are one of the few cases (pardon the pun) where `===` behaves the same as `==`. 

In fact, String `===` is an alias of `==`,  so it is literally the same.
https://github.com/ruby/ruby/blob/v2_7_3/string.c#L3253-L3277

Not so for many other types though.

bxsx: Thanks for the clarification!

Issue #1094, opened 2022-06-27
Ruby-On-Rails SDK - Local variable tracking - Runtime Error


phil-losty-rb:
Instance reported of running into a runtime error trying to enable local variable tracking in Rollbar in our core Ruby on Rails application.

The following steps were used to recreate the issue outside of the client system:

- install the ice_cube gem in a rails application
- Add rollbar to rails (6.1.5.1) application and enable locals
- Create a spec using RSpec (3.11) that calls the following method in production code: 

`IceCube::Schedule.new.occurring_at?(Time.now.utc)`

PR #1095, opened 2022-06-27
(locals) gracefully handle binding error when on a Fiber

waltjones:
### Background

`TracePoint#binding` cannot be called from a Fiber, though I haven't seen anywhere that this is documented. This can be confirmed by reviewing the location where the exception is raised in Ruby:
https://github.com/ruby/ruby/blob/744d17ff6c33b09334508e8110007ea2a82252f5/vm.c#L1240

And of course it can be confirmed by calling `binding` while on a Fiber. The above exception will be produced.

In the reported issue, the error happens when `next` is called on an `Enumerator` instance. Calling `next` causes execution to switch to a Fiber.
https://github.com/ice-cube-ruby/ice_cube/blob/2cdfdbaf75bd809831df1af3f0e96639bbc69ee6/lib/ice_cube/schedule.rb#L217

When local variable capture is enabled in rollbar-gem, the stack frames within the Fiber context are captured. The reported issue happens because `TracePoint#binding` is called while capturing these frames.

### Fix

This PR rescues the failed attempt to get the binding and returns `nil`. Later when building the payload, it skips insertion of locals for frames where the binding is missing.

## Type of change

## Related issues

Fixes: https://github.com/rollbar/rollbar-gem/issues/1094

### Development

- [x] Lint rules pass locally

### Code review

- [x] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [x] "Ready for review" label attached to the PR and reviewers assigned

### DIFF
--- a/lib/rollbar/item/locals.rb
+++ b/lib/rollbar/item/locals.rb
@@ -32,6 +32,8 @@ module Rollbar
         end
 
         def locals_for(frame)
+          return {} unless frame
+
           {}.tap do |hash|
             frame.local_variables.map do |var|
               hash[var] = prepare_value(frame.local_variable_get(var))
--- a/lib/rollbar/notifier/trace_with_bindings.rb
+++ b/lib/rollbar/notifier/trace_with_bindings.rb
@@ -55,12 +55,20 @@ module Rollbar
 
       def frame(trace)
         {
-          :binding => trace.binding,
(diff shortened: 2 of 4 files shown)


cyrusradfar (approved):
Thank you for documenting the changes so clearly

Issue #1106, opened 2022-10-06
Error when sending error with Rack 3.0 app


aglushkov:
Rack 3.0 has breaking change that input is not required to be rewindable 

I have Roda app with Puma and Rack 3.0, and when sending app error it raises rollbar error
```
E, [2022-10-06T23:55:11.483086 #1082613] ERROR -- : [Rollbar] Reporting internal error encountered while sending data to Rollbar.
E, [2022-10-06T23:55:11.498481 #1082613] ERROR -- : [Rollbar] Sending failsafe response due to NoMethodError: "undefined method `rewind' for #<Rack::Lint::Wrapper::InputWrapper:0x00007f605469e8b0 @input=#<Puma::NullIO:0x00007f60535e6d10>>

      rack_req.body.rewind
                   ^^^^^^^" in /home/xxx/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rollbar-3.3.1/lib/rollbar/request_data_extractor.rb:226:in `ensure in rollbar_raw_body_params': build_item in exception_data
I, [2022-10-06T23:55:11.512067 #1082613]  INFO -- : [Rollbar] Sending item
I, [2022-10-06T23:55:11.512332 #1082613]  INFO -- : [Rollbar] Sending json
I, [2022-10-06T23:55:11.993653 #1082613]  INFO -- : [Rollbar] Success
E, [2022-10-06T23:55:11.997402 #1082613] ERROR -- : [Rollbar] Exception: undefined method `rewind' for #<Rack::Lint::Wrapper::InputWrapper:0x00007f605469e8b0 @input=#<Puma::NullIO:0x00007f60535e6d10>>

      rack_req.body.rewind
                   ^^^^^^^
```

PR #1107, opened 2022-10-07
Fix error with body.rewind with rack 3.0

aglushkov:
In rack 3.0 there are no method `#rewind` in Rack::Lint::InputWrapper that existed in rack 2.2

## Type of change

## Related issues
- Fix [#1106](https://github.com/rollbar/rollbar-gem/issues/1106)

### DIFF
--- a/lib/rollbar/request_data_extractor.rb
+++ b/lib/rollbar/request_data_extractor.rb
@@ -214,7 +214,7 @@ module Rollbar
       return {} unless correct_method
       return {} unless json_request?(rack_req)
 
-      raw_body = rack_req.body.read
+      raw_body = read_raw_body(rack_req.body)
       begin
         Rollbar::JSON.load(raw_body)
       rescue StandardError
@@ -222,8 +222,15 @@ module Rollbar
       end
     rescue StandardError
       {}
-    ensure
-      rack_req.body.rewind
+    end
+
+    def read_raw_body(body)
+      return {} unless body.respond_to?(:rewind)
+
+      body.rewind
+      raw_body = body.read
+      body.rewind
+      raw_body
     end
 
     def json_request?(rack_req)
--- a/spec/rollbar/request_data_extractor_spec.rb
+++ b/spec/rollbar/request_data_extractor_spec.rb
@@ -354,6 +354,47 @@ describe Rollbar::RequestDataExtractor do
       end
     end
 
+    context 'with non-rewindable input' do
+      let(:params) { { 'key' => 'value' } }
+      let(:body) { params.to_json }
+      let(:env) do
+        Rack::MockRequest.env_for('/?foo=bar',
+                                  'CONTENT_TYPE' => 'application/json',
+                                  :method => 'POST')
+      end
+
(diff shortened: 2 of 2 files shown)


brianr:
Thanks @aglushkov , we're taking a look at this soon.

waltjones:
@aglushkov Thanks for bringing up this issue, and thank you for the PR.

I looked up the rack issue (https://github.com/rack/rack/issues/1148) and PR (https://github.com/rack/rack/pull/1804) for more insight.

My concern about this PR is that the input is still IO-like, and reading from it will still update the IO cursor. The difference is that since it is non-rewindable, it can only be read once, so rollbar-gem probably shouldn't be reading it. 

Am I missing something that would make this safe as is?

aglushkov:
@waltjones Your concerns are correct. 
- If input was already read,  you can not read it again (so you can receive issues, that input data was not tracked by Rollbar)
- If input was not read, you can read it, but no one can do it after.

There are [Rack::RewindableInput](https://github.com/rack/rack/blob/2.2.0/lib/rack/rewindable_input.rb#L16) class in Rack-2
and additional [Rewindable::Middleware](https://github.com/rack/rack/blob/3.0.0/lib/rack/rewindable_input.rb#L18) in Rack-3 that can be used to get around this issues. But using them can be also an issue, as they will copy input to Tempfile, that can be no desirable behavior for application.

I would suggest using not-public `body.instance_variable_get(:@input)` interface for Rack 3.0 to achieve previous behavior for most applications without Rack::RewindableInput.

```ruby
def rewind_body(rack_req)
     body = rack_req.body
     if body.respond_to?(:rewind)
       body.rewind
     else
       # workaround for Rack 3.0
       # for my application (Rack, Roda, Puma) `body_input` is StringIO, Puma::NullIO, Tempfile, everything is rewindable
       body_input = body.instance_variable_get(:@input) 
       body_input.rewind if body_input.respond_to?(:rewind) # additional check if its not rewindable
   end
end
```

Such should be used before and after reading input.

waltjones:
@aglushkov Nice workaround!

I have two suggestions.
* We need to check for nil `body_input`, since `@input` could be removed/renamed in the future.
* We need to skip `rack_req.body.read` if the input is not rewindable. For example, something like: 
 
```ruby
return {} unless rewindable?(rack_req)
```

aglushkov:
It turns out this `@input` is only present when  `Rack::Lint` middleware is used. It is enabled by default by Rackup, but can be disabled with `--env none` option. In most production apps, where Rollbar is used it should  be disabled, so having
`req.body.rewind if req.body.respond_to?(:rewind)`  should be enough to satisfy new Rack3 specification

waltjones:
@aglushkov That sounds fine. We still need to skip rack_req.body.read if the input is not rewindable.

aglushkov:
@waltjones Sorry for delay, I've update PR, add test

waltjones (approved):
@aglushkov Nice work. Thank you for the test coverage, and thank you again for the PR.

Issue #1126, opened 2023-04-17
Using rescue_from in ApplicationMailer has unexpected results under rails 6 and rollbar 3.4.0


fastjames:
While reviewing dependabot version upgrades recently for a rails app that I maintain, I noticed an unexpected failure in the rollbar update from 3.3.0 to 3.4.0. After investigating further, I narrowed the error source to a use of `rescue_from` in the rails app's ApplicationMailer class to log mail errors in a model attribute. It appears that the new ActiveJob integration may have led to this bug.

I have created a demonstration app visible at https://github.com/fastjames/rollbar_postmark_test that describes the problem. With the rollbar gem version set to 3.4.0, the single spec will fail. If the rollbar gem version is set to 3.3.0 (or 3.3.3), the single spec will pass.

If there is a more recommended way to get what we need with `rescue_from` in the rails app itself, I am happy to open a documentation PR to add it to this gem.

PR #1143, opened 2023-12-27
Fix ActionMailer plugin to add `rescue_from` to the correct class

waltjones:
## Description of the change

This PR has two fixes.
* The `ActiveJob` plugin is now a true plugin, with all of the methods and capabilities that should be present. It now loads at the correct load time, and can be disabled from the Rollbar config. d3e0029f3fbc928adbe44647239dd9e290b1c8e4
* The `rescue_from` handler is added to the correct base class. 4d4f391ee6bd11b96c4fe4acd93337aeaf292634

This PR also refactors the tests to test Rails 6+, which previously didn't get coverage that included the mailer class. Where possible, the tests are now in shared examples called by both the old and new contexts.

Background facts about `rescue_from` handlers.
* These have precedence in the order they are added, with the last added having the highest precedence.
* Only one handler is called for an exception. Handlers are not chained or cascaded.
* The first handler in order of precedence that matches the exception class is the one called.
 
When `MailDeliveryJob` was added in Rails 6.x, it included a `rescue_from` handler [that calls the `rescue_from` of the Mailer class](https://github.com/rails/rails/blob/b1d57fb8c87d771d9a74ac31cd9969e6806b03b2/actionmailer/lib/action_mailer/mail_delivery_job.rb#L40-L46).

When the Rollbar `ActiveJob` plugin installs its handler on `MailDeliveryJob`, it is added after the one in the `MailDeliveryJob` class, and has precedence. This causes the intended handler to never get called, and therefore never pass the error to the Mailer class.

`ActionMailer::Base` is the base class for all Mailers, and this is the correct place to add the Rollbar handler. With this fix, the application's handler will be used if it matches the exception class, and if not, the Rollbar handler is called.

## Type of change

## Related issues

Fixes https://github.com/rollbar/rollbar-gem/issues/1126

### DIFF
--- a/lib/rollbar/plugins/active_job.rb
+++ b/lib/rollbar/plugins/active_job.rb
@@ -27,14 +27,11 @@ Rollbar.plugins.define('active_job') do
 
     if defined?(ActiveSupport) && ActiveSupport.respond_to?(:on_load)
       ActiveSupport.on_load(:action_mailer) do
-        # Automatically add to ActionMailer::DeliveryJob
-        if defined?(ActionMailer::DeliveryJob)
+        if defined?(ActionMailer::MailDeliveryJob) # Rails >= 6.0
+          ActionMailer::Base.send(:include, Rollbar::ActiveJob)
+        elsif defined?(ActionMailer::DeliveryJob) # Rails < 6.0
           ActionMailer::DeliveryJob.send(:include,
                                          Rollbar::ActiveJob)
-        end
-        # Rails >= 6.0
-        if defined?(ActionMailer::Base)
-          ActionMailer::Base.send(:include, Rollbar::ActiveJob)
         end
       end
     end
--- a/lib/rollbar/plugins/active_job.rb
+++ b/lib/rollbar/plugins/active_job.rb
@@ -33,8 +33,8 @@ Rollbar.plugins.define('active_job') do
                                          Rollbar::ActiveJob)
         end
         # Rails >= 6.0
-        if defined?(ActionMailer::MailDeliveryJob)
-          ActionMailer::MailDeliveryJob.send(:include, Rollbar::ActiveJob)
+        if defined?(ActionMailer::Base)
+          ActionMailer::Base.send(:include, Rollbar::ActiveJob)
         end
       end
     end
--- a/lib/rollbar/plugins/active_job.rb
+++ b/lib/rollbar/plugins/active_job.rb
@@ -14,11 +14,16 @@ Rollbar.plugins.define('active_job') do
                      arguments
                    end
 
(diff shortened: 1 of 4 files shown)


mudetroit (approved):
Makes sense to me, thanks for the description in the PR help clear up the context

lib/rollbar/plugins/active_job.rb
```diff
@@ -1,35 +1,44 @@
...
-                      :arguments => args)
-        raise exception
+  execute do
+    module Rollbar
+      # Report any uncaught errors in a job to Rollbar and reraise
+      module ActiveJob
+        def self.included(base)
+          base.send :rescue_from, Exception do |exception|
+            args = if self.class.respond_to?(:log_arguments?) && !self.class.log_arguments?
+                     arguments.map(&Rollbar::Scrubbers.method(:scrub_value))
```
lucasmazza, line 12: Hey @waltjones, I noticed that the rollbar update to 3.5.0 broke one of the mailers of our app and tracked it down here: `ActionMailer::Base` doesn't have an `arguments` method and `NameError: undefined local variable or method 'arguments' for #<YourMailerClass:0x0000000036ada8>>` here

I assume this was fixed on the tests here with the `attr_accessor :arguments` on the TestMailer, but that doesn't happen on a regular rails app

lib/rollbar/plugins/active_job.rb
```diff
@@ -1,35 +1,44 @@
...
-  # Report any uncaught errors in a job to Rollbar and reraise
-  module ActiveJob
-    def self.included(base)
-      base.send :rescue_from, Exception do |exception|
-        args = if self.class.respond_to?(:log_arguments?) && !self.class.log_arguments?
-                 arguments.map(&Rollbar::Scrubbers.method(:scrub_value))
-               else
-                 arguments
-               end
+Rollbar.plugins.define('active_job') do
```
adriancb, line 1: 👋 - this change produces:

```
NameError: uninitialized constant Rollbar::ActiveJob (NameError)

  include Rollbar::ActiveJob
                 ^^^^^^^^^^^
```
which means that https://docs.rollbar.com/docs/activejob-integration is no longer true. Is there a change required to the documentation?

https://docs.rollbar.com/docs/gem-plugins is unclear on how plugins are loaded, so we were unable to experiment with a solution.

lib/rollbar/plugins/active_job.rb
waltjones: @lucasmazza Fix is released in 3.5.1. https://github.com/rollbar/rollbar-gem/releases/tag/v3.5.1

Issue #1145, opened 2024-01-10
NameError: undefined local variable or method `arguments' for an instance of xxxMailer


nov:
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/plugins/active_job.rb is expecting `ActionMailer::Base` children have `arguments`, but actually not.
thus causing the error in title

womanonrails:
I have the same problem `rollbar-gem` version `3.5.0`.

waltjones:
Thank you for the reports. Working to get a fix out today.

waltjones:
Released in 3.5.1. https://github.com/rollbar/rollbar-gem/releases/tag/v3.5.1

isaacbowen:
Thanks @waltjones! :)

PR #1146, opened 2024-01-11
fix: compatibility fixes for ActionMailer plugin

waltjones:
## Description of the change

This PR fixes several compatibility issues with plugin support for ActionMailer.
* Correctly detect whether included in a mailer class.
* Report metadata available in subclasses of ActionMailer::Base, and don't reference attributes that aren't. (e.g. `arguments`)
* Prevent double Rollbar reports, caused by the `rescue_from` handler getting called twice.

## Type of change

## Related issues

Fixes: https://github.com/rollbar/rollbar-gem/issues/1145

### DIFF
--- a/lib/rollbar/plugins/active_job.rb
+++ b/lib/rollbar/plugins/active_job.rb
@@ -8,22 +8,32 @@ Rollbar.plugins.define('active_job') do
       module ActiveJob
         def self.included(base)
           base.send :rescue_from, Exception do |exception|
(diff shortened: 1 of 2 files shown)


matux (approved):
Looks good

lucasmazza:
I tested this with our app that has a test that fails with 3.5.0 and everything looks good. Thanks for fixing this ❤️

Issue #1105, opened 2022-09-16
Use The new Rails error reporter


morgoth:
The built-in error reporter is now available in Rails.
https://edgeguides.rubyonrails.org/error_reporting.html

It should simplify and cleanup the integration for the rollbar gem.
There is already an error tracker that is integrated with it, vide https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/error_subscriber.rb

waltjones:
@morgoth thank you! Support for this is published in v3.6.0. https://github.com/rollbar/rollbar-gem/releases/tag/v3.6.0

Docs will be updated. See the PR for config.

Issue #1156, opened 2024-05-03
Rollbar::Logger not working well with the newest logger gem


randypuro:
Rollbar::Logger inherits from ::Logger, but [doesn't call it's constructor](https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/logger.rb#L25).

logger gem v1.6.0 (which, for instance, comes with ruby 3.3.1) [initializes other parameters in that constructor](https://github.com/ruby/logger/blob/master/lib/logger.rb#L587) which are used in other methods (like [the level accessor](https://github.com/ruby/logger/blob/master/lib/logger.rb#L384)).

It means accessing the `level` in Rollbar::Logger raises an exception. Rollbar::Logger should probably also use the level accessor (instead of the instance variable) to be more resilient to these kinds of improvements/changes to the base ::Logger.

And while I understand the point of raising exceptions on formatter, etc; it means the Logger doesn't play well with the newer Rails logger broadcast functionality which tries to copy formatters down to any new broadcast children.

waltjones:
@randypuro Thank you for the report!

Published in v3.6.0. https://github.com/rollbar/rollbar-gem/releases/tag/v3.6.0

Issue #1153, opened 2024-03-20
Using ActiveSupport::ErrorReporter


olegantonyan:
I'm trying to make Rollbar work with Rails' new error handling mechanism https://guides.rubyonrails.org/error_reporting.html 
```
::Rollbar.configure do |config|
  config.access_token = '...'
  config.environment = ::Rails.env.to_s

  config.disable_monkey_patch = true
  config.disable_rack_monkey_patch = true
  config.disable_action_mailer_monkey_patch = true
end

class ErrorSubscriber
  def report(error, handled:, severity:, context:, source: nil)
    ::Rollbar.log(severity.to_s, error, context: context, source: source)
  end
end

::Rails.error.subscribe(ErrorSubscriber.new)
```
But the problem is Rollbar monkey patches everything to intercept errors, so it ends up in duplicate error reports: one from Rails ErrorSubscriber, and another from within Rollbar gem. 

My first attempt to solve this was trying to disable all monkey patches. But no luck `disable_monkey_patch` config option does not disable Rails (`lib/rollbar/plugins/rails.rb`). Adding `dependency { !configuration.disable_monkey_patch }` there doesn't work. Only completly commenting out all monkey patch code works.

Eventally what I'd like to rely on Rails' ErrorSubscriber alone, and use only Rollbar's server-side to receive messages. No monkey patches, no error interceptions - nothing. Maybe even use raw api directly without the gem (https://docs.rollbar.com/reference/create-item)

This may sound kind of stupid, but the goal is to test how this new ErrorSubscriber works. Because "traditional" approach with monkey-patches works only until you have multiple gems monkey-patching the same thing has already shot me in the foot, so the addition of Rails native error handling mechanism seems like a good idea.

tmimura39:
> But the problem is Rollbar monkey patches everything to intercept errors, so it ends up in duplicate error reports: one from Rails ErrorSubscriber, and another from within Rollbar gem.

This issue is important to me as well.

The [sentry gem](https://github.com/getsentry/sentry-ruby) has taken the following actions.
https://github.com/getsentry/sentry-ruby/pull/1738

If the Rollbar Gem does not support this, we have no choice but to “not use ActiveSupport::ErrorReporter” or “create our own thin Rollbar Middrware”.

It would be helpful if you could give us Rollbar official thoughts.

waltjones:
[...]

PR #1161, opened 2024-08-26
Enable Rails 7 error reporter via Rollbar config

waltjones:
## Description of the change

Optionally enables the Rails 7 error reporter. https://edgeguides.rubyonrails.org/error_reporting.html

The new Boolean config key is `enable_rails_error_subscriber`.
```ruby
      Rollbar.configure do |config|
        config.enable_rails_error_subscriber = true
      end
```

This is off by default for back compatibility, and can be enabled in the Rollbar initializer, or enabled/disabled at any time via the `Rollbar.configure` method.

This is available with Rails 7.1+ because that's when `unsubscribe` was introduced.

This PR also fixes an issue with the `Rollbar::Logger` class.
* `initialize` should call `super`. This issue breaks in recent Logger versions.
* The Logger version should be used to branch in tests.

Normally I'd put this in a separate PR, but since it breaks CI I've added it here.

Note: This PR also enables both hash syntaxes in rubocop. 

## Type of change

- [x] New feature (non-breaking change that adds functionality)

## Related issues

Fixes:
https://github.com/rollbar/rollbar-gem/issues/1105
https://github.com/rollbar/rollbar-gem/issues/1153
https://github.com/rollbar/rollbar-gem/issues/1156

### Development

- [x] Lint rules pass locally

### Code review

- [x] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [x] Changes have been reviewed by at least one other engineer

### DIFF
--- a/lib/rollbar/exception_reporter.rb
+++ b/lib/rollbar/exception_reporter.rb
@@ -27,7 +27,8 @@ module Rollbar
     end
 
     def capture_uncaught?
-      Rollbar.configuration.capture_uncaught != false
+      Rollbar.configuration.capture_uncaught != false &&
+        !Rollbar.configuration.enable_rails_error_subscriber
     end
 
     def log_exception_message(exception)
--- a/lib/rollbar/plugins/rails/error_subscriber.rb
+++ b/lib/rollbar/plugins/rails/error_subscriber.rb
@@ -1,0 +1,9 @@
+module Rollbar
+  class ErrorSubscriber
+    def report(error, handled:, severity:, context:, source: nil)
+      extra = context.is_a?(Hash) ? context.deep_dup : {}
+      extra[:custom_data_method_context] = source
+      Rollbar.log(severity, error, extra)
+    end
+  end
+end
--- a/lib/rollbar/plugins/rails/error_subscriber.rb
+++ b/lib/rollbar/plugins/rails/error_subscriber.rb
@@ -1,6 +1,9 @@
 module Rollbar
   class ErrorSubscriber
     def report(error, handled:, severity:, context:, source: nil)
+      # The default `nil` for capture_uncaught means `true`. so check for false.
+      return unless handled || Rollbar.configuration.capture_uncaught != false
+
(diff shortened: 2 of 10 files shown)


lib/rollbar/exception_reporter.rb
```diff
@@ -27,7 +27,8 @@ def report_exception_to_rollbar(env, exception)
     end
 
     def capture_uncaught?
-      Rollbar.configuration.capture_uncaught != false
+      Rollbar.configuration.capture_uncaught != false &&
+        !Rollbar.configuration.enable_rails_error_subscriber
```
matux, line 31: If we are subscribed to rails' error subscriber, our `capture_uncaught` setting becomes redundant?

lib/rollbar/plugins/rails/error_subscriber.rb
```diff
@@ -0,0 +1,9 @@
+module Rollbar
+  class ErrorSubscriber
+    def report(error, handled:, severity:, context:, source: nil)
+      extra = context.is_a?(Hash) ? context.deep_dup : {}
+      extra[:custom_data_method_context] = source
+      Rollbar.log(severity, error, extra)
+    end
```
matux, line 7: `handled:` is unused?

matux (commented):
Just two small comments

lib/rollbar/exception_reporter.rb
waltjones: This prevents double sends, because the Rails error reporter (not our subscriber) re-raises in some cases.

lib/rollbar/plugins/rails/error_subscriber.rb
waltjones: Yes, though I'm considering using the `capture_uncaught` flag here to swallow the error when `!handled && !capture_uncaught`.

matux (approved):
Looks awesome!

dcrec1:
When used with Sidekiq, this reports a lot of `Sidekiq::JobRetry::Handled` errors that can't be filtered. Is this expected?

PR #1199, opened 2026-01-08
Fix JSON::GeneratorError for objects with invalid UTF-8 in to_s

### LINKED ISSUE #1196: Internal failsafe when serializing invalid utf8 content
We're seeing this error pop up in production every now and then that we unfortunately can't reliably reproduce:

```
Failsafe from rollbar-gem. JSON::GeneratorError: 
"source sequence is illegal/malformed utf-8" in /usr/local/bundle/ruby/3.4.0/gems/json-2.15.1/lib/json/common.rb:448:in 'JSON::Ext::Generator::State.generate': build_item in exception_data
```

There's a corresponding log line that suggests the cause:
```
2025-12-03T12:30:08.527Z
[b730d3f1-fc6a-8a00-3b27-70dd6a5e177a] ActionController::BadRequest (Invalid request parameters: Invalid encoding for parameter: redacted�redacted.pdf)
2025-12-03T12:30:08.527Z
[b730d3f1-fc6a-8a00-3b27-70dd6a5e177a] Caused by: Rack::QueryParser::InvalidParameterError (Invalid encoding for parameter: redacted�redacted.pdf)
```

What I think is happening:
1) the request triggers an error
2) rollbar tries to report that error, encounters an unknown internal error
3) rollbar tries to report that internal error, but hits JSON::GeneratorError when building the item to report

While exploring possible causes, I found a reliable way to circumvent `Rollbar::Encoding.encode` - put anywhere in the serializable context an item that has a `to_s` method that returns invalid utf-8. Here's a failing spec in `item_spec.rb`:

```ruby
    context 'with object whose to_s returns invalid UTF-8' do
      let(:bad_object) do
        obj = Object.new
        def obj.to_s
          "Invalid: test\xFF\xFEbad".force_encoding('UTF-8')
        end
        obj
      end
      let(:item) do
        Rollbar::Item.new(
          :level => 'error',
          :extra => { :bad => bad_object },
          :configuration => configuration,
          :logger => logger,
          :scope => {},
          :notifier => notifier
        ).tap(&:build)
      end

      it 'encodes the object to valid UTF-8 and serializes without error' do
        json = nil
        expect { json = item.dump }.not_to raise_error

        expect(json).to be_a(String)
        expect(json).to include('Invalid:')
      end
    end
```

I also have a potential fix for that:

```
diff --git a/lib/rollbar/encoding.rb b/lib/rollbar/encoding.rb
index 8931c4f..2c62294 100644
--- a/lib/rollbar/encoding.rb
+++ b/lib/rollbar/encoding.rb
@@ -10,11 +10,16 @@ module Rollbar
     end
 
     def self.encode(object)
-      can_be_encoded = object.is_a?(String) || object.is_a?(Symbol)
-
-      return object unless can_be_encoded
-
-      encoding_class.new(object).encode
+      case object
+      when String, Symbol
+        encoding_class.new(object).encode
+      when Numeric, TrueClass, FalseClass, NilClass
+        object # JSON-native types, safe to pass through
+      else
+        # Other objects will have to_s/to_json called during JSON serialization
+        # Encode their string representation to ensure valid UTF-8
+        encoding_class.new(object.to_s).encode
+      end
     end
   end
 end
```

I'm opening an issue instead of a PR because I'm still not 100% sure this is the root cause, and I can't find a reliable end-to-end reproduction.

grk:
## Description of the change

Rollbar::Encoding.encode only processed String and Symbol types. When objects with to_s methods returning invalid UTF-8 were included in the payload, they bypassed encoding and caused JSON serialization to fail with "source sequence is illegal/malformed utf-8".

Now all non-JSON-native types are passed through the Encoder, which already calls to_s and handles encoding normalization.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fixes #1196

### Development

- [ ] Lint rules pass locally

### Code review

- [ ] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers assigned
- [ ] Issue from task tracker has a link to this pull request

### DIFF
--- a/lib/rollbar/encoding.rb
+++ b/lib/rollbar/encoding.rb
@@ -10,11 +10,12 @@ module Rollbar
     end
 
     def self.encode(object)
-      can_be_encoded = object.is_a?(String) || object.is_a?(Symbol)
-
-      return object unless can_be_encoded
-
-      encoding_class.new(object).encode
+      case object
+      when Numeric, TrueClass, FalseClass, NilClass
+        object
+      else
+        encoding_class.new(object).encode
+      end
     end
   end
 end
--- a/spec/rollbar/util_spec.rb
+++ b/spec/rollbar/util_spec.rb
@@ -134,10 +134,14 @@ describe Rollbar::Util do
     it 'should replace invalid utf8 values' do
       bad_key = force_to_ascii("inner \x92bad key")
 
+      class ObjClass; def to_s; "bad obj\255".force_encoding('ASCII-8BIT'); end; end
+      bad_obj = ObjClass.new
+
       payload = {
         :bad_value => force_to_ascii("bad value 1\255"),
         :bad_value2 => force_to_ascii("bad\255 value 2"),
         force_to_ascii("bad\255 key") => 'good value',
+        :bad_obj => bad_obj,
         :hash => {
           :inner_bad_value => force_to_ascii("\255\255bad value 3"),
           bad_key.to_sym => 'inner good value',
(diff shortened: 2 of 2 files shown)


waltjones (approved):
Looks good , and thank you!

waltjones:
Released in v3.8.0.

PR #1208, opened 2026-08-25
Add config.js_exclude_paths to skip rollbar.js injection on some paths

bugthing:
`Rollbar::Middleware::Js` currently injects the rollbar.js snippet into every `200` HTML response app-wide, with no way to opt individual paths out.

The poblem is for routes that serve their own restrictive `Content-Security-Policy`, for example `rswag-ui`'s Swagger UI page sets a hardcoded, narrow CSP (`script-src 'self' 'unsafe-inline'`) that doesn't allow `cdn.rollbar.com`. The injected snippet's breaks their CSP

## Change

Adds `config.js_exclude_paths`, an array of `String` (matched as a `PATH_INFO` prefix) or `Regexp` entries. When a request's path matches any entry, `Rollbar::Middleware::Js` skips injecting the config/snippet `<script>` tags for that response, using the same short-circuit as the existing `enabled?`/`attachment?`/`streaming?` checks in `add_js?`.

```ruby
Rollbar.configure do |config|
  config.js_enabled = true
  config.js_exclude_paths = ['/api-docs', %r{\A/health}]
end
```

Defaults to `[]`, so this is fully backward compatible — no behavior change unless the option is explicitly set.


lib/rollbar/middleware/js.rb
```diff
@@ -46,7 +46,18 @@ def enabled?
...
+          !excluded_path?(env)
+      end
+
+      def excluded_path?(env)
+        exclude_paths = config[:exclude_paths]
+        return false unless exclude_paths
+
+        path = env['PATH_INFO'].to_s
+        exclude_paths.any? do |matcher|
+          matcher.is_a?(Regexp) ? matcher.match?(path) : path.start_with?(matcher.to_s)
```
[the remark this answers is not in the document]
bugthing: Fixed in 13791d3 — switched to `Regexp#match` which is available on all Ruby versions declared by the gemspec.

lib/rollbar/middleware/js.rb
```diff
@@ -46,7 +46,18 @@ def enabled?
...
-          html?(headers) && !attachment?(headers) && !streaming?(env)
+          html?(headers) && !attachment?(headers) && !streaming?(env) &&
+          !excluded_path?(env)
+      end
+
+      def excluded_path?(env)
+        exclude_paths = config[:exclude_paths]
+        return false unless exclude_paths
+
+        path = env['PATH_INFO'].to_s
```
zdavis-rollbar: This one doesn't catch the rswag-ui page. We ran it with `exclude_paths` set to `['/api-docs']` and requested `/api-docs/index.html`. By the time this check runs, Rails has already changed the path to `/index.html` and moved the `/api-docs` part into SCRIPT_NAME. Rails only puts the path back if the engine passes on the request, and the Swagger page doesn't, so the match fails and the snippet still goes in. You can see it in `recognize` in `action_dispatch/journey/router.rb`.

The specs don't catch it because they pass in a fixed path and never go through the router.

I think the fix is to grab the path before calling the app, as SCRIPT_NAME plus PATH_INFO. That does change what the setting matches: it becomes the full path someone sees in the browser, including the sub-path if the app is hosted under one, and the description currently says PATH_INFO. I think that's the better behavior, but let's put it in the docs. Can we also add a test that goes through a real mounted engine so this doesn't come back?

lib/rollbar/middleware/js.rb
```diff
@@ -46,7 +46,18 @@ def enabled?
...
+        exclude_paths.any? do |matcher|
```
zdavis-rollbar: I know the description says to pass an array, but passing a single string by mistake takes rollbar.js off every page in the app. `.any?` raises NoMethodError on a string, the rescue in `#call` logs "Rollbar.js could not be added" and moves on, and that happens on every HTML response, including pages that aren't excluded. Wrapping it in `Array(exclude_paths)` makes that mistake safe and lets someone pass a single Regexp too.

lib/rollbar/middleware/js.rb
```diff
@@ -46,7 +46,18 @@ def enabled?
...
+          matcher.is_a?(Regexp) ? matcher.match(path) : path.start_with?(matcher.to_s)
```
zdavis-rollbar: If an entry is nil or an empty string, it matches every page, because `nil.to_s` is `''` and every path starts with `''`. Someone who writes `[ENV['ROLLBAR_JS_SKIP']]` and forgets to set the variable would turn rollbar.js off across the whole app, with nothing in the logs to tell them. Can we skip nil and empty entries?

Smaller question on the same line. `'/api-docs'` also matches `/api-docs-internal`. That's how prefix matching works and it's what the description says, so I don't think it's wrong, but I wonder if matching the exact path or anything under it would surprise people less. If you'd rather keep prefix matching, a note in the docs that `%r{\A/api-docs(?:/|\z)}` gets you the stricter version would cover it.

zdavis-rollbar (changes requested):
Thanks for putting this together, Ben.  I left three comments on the matching code, and the first one is one that needs to be addressed. We ran it with a mounted engine and the snippet still gets injected on the Swagger page, so as written this doesn't fix the case you opened it for.

Can you also add the new option to `docs/configuration.md` so people can find it? Once those are sorted out, I'm happy to get this merged.

lib/rollbar/middleware/js.rb
bugthing: wow!, that is a great find, thanks for the clear explanation too. I will get this issue addressed (likely in the manor you suggest)

lib/rollbar/middleware/js.rb
bugthing: great point. I like that Array cast niceity :)

lib/rollbar/middleware/js.rb
bugthing: hmm.. I think you're right. I'll look into tweaking it to match a path and anything under would make sense. Althou it might be more complex if one considers a single root level page :thinking:

